PR #1, opened 2019-09-16
Feature/add verifications

alissonfpmorais:
This PR add new types to retrieve verifications from the API.
Below are the changelog:
- [x] add query to get an specific verification;
- [x] add query to get all verifications from an specific validator;
- [x] add resolvers for both verification and verifications;
- [x] add types related to a verification;
- [x] add types related to a verification_history;
- [x] remove `BeSpiral` from type description;
- [x] add tests.

### DIFF
--- a/lib/bespiral_web/schema/commmune_types.ex
+++ b/lib/bespiral_web/schema/commmune_types.ex
@@ -8,15 +8,15 @@ defmodule BeSpiralWeb.Schema.CommuneTypes do
   import Absinthe.Resolution.Helpers, only: [dataloader: 1]
   alias BeSpiralWeb.Resolvers.Commune
 
-  @desc "Community Queries on BeSpiral"
+  @desc "Community Queries"
   object :community_queries do
-    @desc "A list of sales in BeSpiral"
+    @desc "A list of sales"
     field :sales, non_null(list_of(non_null(:sale))) do
       arg(:input, non_null(:sales_input))
       resolve(&Commune.get_sales/3)
     end
 
-    @desc "A list of communities in BeSpiral"
+    @desc "A list of communities"
     field :communities, non_null(list_of(non_null(:community))) do
       resolve(&Commune.get_communities/3)
     end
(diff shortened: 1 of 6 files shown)


lib/bespiral_web/schema/commmune_types.ex
```diff
@@ -99,7 +111,18 @@ defmodule BeSpiralWeb.Schema.CommuneTypes do
...
-  @desc "A community on BeSpiral"
+  @desc "Input to collect verifications"
+  input_object :verifications_input do
+    field(:validator, non_null(:string))
+  end
+
+  @desc "Input to collect verification"
+  input_object :verification_input do
+    field(:id, non_null(:integer))
+    field(:validator, non_null(:string))
```
zacck: @alissonfpmorais  if I have an id then I shouldn't really need a validator or is this for a compound key?

lib/bespiral_web/schema/commmune_types.ex
```diff
@@ -274,9 +297,50 @@ defmodule BeSpiralWeb.Schema.CommuneTypes do
     field(:created_at, non_null(:datetime))
   end
 
+  @desc "A minimal verification used to on a historical list"
+  object :verification_history do
```
zacck: @alissonfpmorais where is this used it looks like we are building data for the client that can be picked using GraphQL?

lib/bespiral_web/schema/commmune_types.ex
```diff
@@ -274,9 +297,50 @@ defmodule BeSpiralWeb.Schema.CommuneTypes do
...
+    field(:objective_id, non_null(:integer))
+    field(:action_id, non_null(:integer))
+    field(:claim_id, non_null(:integer))
+    field(:logo, non_null(:string))
+    field(:description, non_null(:string))
+    field(:created_at, non_null(:datetime))
+    field(:status, non_null(:verification_history_status_type))
+  end
+
+  @desc "A claimed action verification"
```
zacck: Same here

lib/bespiral_web/schema/commmune_types.ex
alissonfpmorais: @zacck this is used liked a compound key

lib/bespiral_web/schema/commmune_types.ex
alissonfpmorais: @zacck right now we can get this data from the GraphQL, but after getting it the frontend also have to do some computations before using it. This new type enable the frontend to use the data the way it receives.

lib/bespiral_web/schema/commmune_types.ex
alissonfpmorais: Same as comment above

alissonfpmorais:
> Hey @alissonfpmorais I have a few questions and concerns over the context code for this but first I would like to know where are these types used in the front end?

These types are going to be used in [these screens](https://www.figma.com/file/H8Hxt5toqohyvEi57J27rf/ClaimNewelements?node-id=0%3A1).
Another point is that the entire `Activity` module is going to be removed and replaced by a single query to `verifications_history`.

zacck:
Hi @alissonfpmorais I am wondering how come we have to do this server side as well? I think the Activity module covers it well enough and avoid super specific code on the server side. As in it uses the GraphQl interface almost exactly as it is meant ?

PR #3, opened 2019-09-20
Feature: validator claim query

zacck:
## Changes Proposed 

- [x]  Collect all claims for a validator
- [x] Expose query to fetch claims
- [x] Ensure to collect only claims I can vote or I have voted on


lib/bespiral/commune/commune.ex.orig
```diff
@@ -0,0 +1,460 @@
+defmodule BeSpiral.Commune do
```
lucca65: maybe this is leftoff from vim?

priv/repo/migrations/20190920061915_verification_notification.exs
```diff
@@ -1,45 +0,0 @@
...
-      FOR EACH ROW
-      EXECUTE PROCEDURE #{@function}()
-    """)
-  end
-
-  def down do
-    execute("DROP FUNCTION IF EXISTS #{@function} CASCADE")
-    execute("DROP TRIGGER IF EXISTS #{@event} ON #{@table}")
-  end
-end
```
lucca65: 100% we don't already migrated this in some database?

lib/bespiral/commune/commune.ex.orig
zacck: yeah left overs

priv/repo/migrations/20190920061915_verification_notification.exs
zacck: @lucca65  even if we have the new migrations have new timestamps they will be run again

lib/bespiral/commune/commune.ex
```diff
@@ -89,6 +90,40 @@ defmodule BeSpiral.Commune do
...
+        where: a.is_completed == false,
+        join: v in Validator,
+        on: v.action_id == a.id and v.validator_id == ^account,
+        join: c in Claim,
+        on: c.action_id == a.id,
+        select: c
+      )
+      |> Repo.all()
+
+    voted_claims =
```
alissonfpmorais: Since the `available_claims` is all claims from a specific validator and `voted_claims` are all voted claims from a specific validator, every voted claim will be duplicated in the GraphQL response

lib/bespiral_web/schema/commmune_types.ex
```diff
@@ -63,6 +69,11 @@ defmodule BeSpiralWeb.Schema.CommuneTypes do
     end
   end
 
+  @desc "Input to collect claims"
+  input_object :claims_input do
+    field(:validator, :string)
```
alissonfpmorais: I believe that this type must be `non_null(:string)` or you should check if `validator` exists before using it

lib/bespiral/commune/commune.ex
zacck: Hmm @alissonfpmorais  my tests seem to tell me otherwise, but then again I wrote them for this scenario. Would you have a dataset that shows this please?

alissonfpmorais: Sure! I sent you on slack

priv/repo/migrations/20190920061915_verification_notification.exs
lucca65: ok!

lib/bespiral_web/schema/commmune_types.ex
lucca65: 🤔 interesting, it makes sense.

zacck: Yes I agree

lib/bespiral/commune/commune.ex
zacck: Hey @alissonfpmorais  I need that dataset likely in csv please the sql file you sent me is incompatible

zacck: @alissonfpmorais  I managed to replicate the issue in my tests and found a solution for this, it should work now

lib/bespiral_web/schema/commmune_types.ex
```diff
@@ -37,6 +37,12 @@ defmodule BeSpiralWeb.Schema.CommuneTypes do
       arg(:input, non_null(:sale_input))
       resolve(&Commune.get_sale/3)
     end
+
+    @desc "A list of claims"
+    field :claims, list_of(non_null(:claim)) do
```
alissonfpmorais: I think that this is going to be the same as the input: `non_null(list_of(non_null(:claim)))`.
This way we can avoid the type `Maybe (List Claim)` in Elm

zacck: Ah, hmmm ok since the query will return an empty list anyway

priv/repo/migrations/20190923130436_verification_function.exs
```diff
@@ -8,6 +8,9 @@ defmodule BeSpiral.Repo.Migrations.VerificationFunction do
   @event "claims_changed"
 
   def up do
+    # Remove botched former migration
```
lucca65: botched?

zacck: Added that to ensure that the other migration is removed so this can be added

zacck:
@alissonfpmorais  I think that should make it work, please see if it agrees with yours

priv/repo/migrations/20190923130436_verification_function.exs
zacck: You know too many english words and I too few portuguese words

lucca65: lets say I much rather learn Swahili words 👀

zacck: Pole pole, utaelewa

lucca65: irei meu caro, depois do 日本語 será minha próxima!

lucca65: actually just started! there is Swahili in duolingo!!! 🎉 

> wewe ni  Mkenya!

alissonfpmorais:
> @alissonfpmorais I think that should make it work, please see if it agrees with yours

Hey @zacck I've tested and now looks like that the query return the correct claim, but before merge it, I would like to know from you @zacck and @lucca65 how we gonna use those queries.

The dashboard also needs data related to community (`symbol` and `logo` related to each claim) and objective (`id`, also related to each claim). In that case the frontend will do multiples queries and then join all the results?
The same happens to verification screen, not only we need community and objective data, but, in case of multiples queries, we also need a way to get an specific entry of objective and claim.

Is this right? What you folks think about that?

lucca65:
yeah this is a possibility, but we can also query several different queries on GraphQL, no need to actually do different queries, for example:

```
query {
  x: community(symbol:"BES") {
    description
  }
  y: community(symbol:"LOR") {
    description
  }
}

```

Maybe to improve this conversation, can you please send me the graphql query to the claims, and we can create a new one that contemplates the community queries as well so you get descriptions, logo and symbols to your heart's contents

zacck:
@alissonfpmorais  @lucca65 with a couple of modifications to our Schema,

That is, adding the `objective` association to an `action` and `community` association to an `objective` we can pick this data entirely from the claims query as follows 

```
query ($input: ClaimsInput!) {
  claims(input: $input) {
    id
    createdAt
    action {
      id
      objective {
        id
        community {
          symbol
          logo
        }
      }
    }
  }
}
```

alissonfpmorais:
> @alissonfpmorais @lucca65 with a couple of modifications to our Schema,
> 
> That is, adding the `objective` association to an `action` and `community` association to an `objective` we can pick this data entirely from the claims query as follows
> 
> ```graphql
> query ($input: ClaimsInput!) {
>   claims(input: $input) {
>     id
>     action {
>       objective {
>          id
>         community {
>           symbol
>           logo
>         }
>       }
>     }
>   }
> }
> ```

@zacck Really cool dude, like it a lot!!

One more thing is that we also need a way to get a specific verification for the verification screen. Maybe something like?

```graphql
query($input: ClaimInput!) { // with ClaimInput being claim ID and the validator who is requiring the claim (so you can check if that validator can verify that claim or not)
  claim(input: $input) {
     // data needed
  }
}
```

zacck:
@alissonfpmorais  I can do the query you need `(claim(claim_input))` however I don't see the need to require the validator since the backend doesn't need to know about permissions really and the front end will have the validator id which it can tell from eos ... what do you think

alissonfpmorais:
> @alissonfpmorais I can do the query you need `(claim(claim_input))` however I don't see the need to require the validator since the backend doesn't need to know about permissions really and the front end will have the validator id which it can tell from eos whether they can verify or not ... what do you think

Yeah, since I've those data I can check it on frontend.
Makes sense! It's fine for me dude

zacck:
@alissonfpmorais  done 👍

PR #11, opened 2019-10-22
Add objective query

lucca65:


zacck:
@lucca65 In essence this introduces ease of auto generation of code as we do with the `elm-graphql` library and as we go along and as our inputs grow its easier to modify an input object rather than building up complex arguments. 

It's one of those things I have picked up working with multiple graphql APIs let me know if I can send you more texts to justify if you need ...

PR #15, opened 2019-11-12
Feature/invitations

lucca65:
Handle invitation creation and mailing


lib/bespiral/commune/commune.ex
```diff
@@ -210,8 +210,8 @@ defmodule BeSpiral.Commune do
   """
   @spec get_validator_claims(String.t()) :: {:ok, list(Claim.t())} | {:error, term}
   def get_validator_claims(account) do
-    available_claims =
```
zacck: credo? Is this the recommended way?

zacck: *whitespace ... Ehi!*

PR #22, opened 2019-12-10
Add to and from profiles to account to the transfer object

zacck:
## Changes proposed

- [x] Expose from and to profile in transfers for easier query


test/bespiral_web/schema/resolvers/commune_test.exs
```diff
@@ -337,6 +337,7 @@ defmodule BeSpiralWeb.Schema.Resolvers.CommuneTest do
                end)
     end
 
+    @tag :trans
```
lucca65: why use this tag? maybe we can just write down `transfer`

zacck:
Lol @lucca65  thats a left over ...

PR #43, opened 2020-05-29
Add GraphQL API for Payment History

amiskov:
This is the backend part of a [Payment History](https://github.com/cambiatus/frontend/issues/251) page.

## Changes Proposed (a list of new changes introduced by this PR)
Add query and related types for fetching transfers to a user (recipient) from other users (payers) with the ability to filter transfers by date and payer.

## How to test (a list of instructions on how to test this PR)
The query should return the actual data:

```graphql
query ($input: PaymentHistoryInput!, $num:Int!) {
  paymentHistory(input: $input, first: $num) {
    edges {
      node {
        from {
          account
        }
        to {
          account
        }
        createdAt
      }
    }
  }
}
```

Variables:

```graphql
{
  "input": {
    "recipient": "lucca",
    "payer": "bespiral",
    "date": "2020-04-28"
  },
  "num": 10
}
```


lib/cambiatus/accounts/accounts.ex
```diff
@@ -22,6 +24,72 @@ defmodule Cambiatus.Accounts do
     end
   end
 
+  @doc "Fetch the list of payers filtered by the given account name"
+  @spec filter_payers_by_account(String.t(), String.t()) :: {:ok, list(string)}
+  def filter_payers_by_account(recipient, payer) do
```
lucca65, line 29: I thinking that maybe its best to use something like `get_transfers` but with an extra param. Using a function named filter make it looks like its a filter, but its a function that get transfers, but with a certain filter

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -15,6 +15,25 @@ defmodule CambiatusWeb.Schema.AccountTypes do
       arg(:input, non_null(:profile_input))
       resolve(&Accounts.get_profile/3)
     end
+
+    @desc "List of payers to the given recipient filtered by the part of the account name."
+    field :get_payers, list_of(:profile) do
```
lucca65, line 20: I feel again that we should just make sure we are getting transfers, but filtering it by a given context

lib/cambiatus/accounts/accounts.ex
amiskov: Now I see that I decided to make a separate function because of the pagination of the Transfer Connection. But it seems like it's just a `|> get_transfers_from(pagination_args)` line and I can add more `get_transfers` function clauses without it.

`get_transfers` may become a really big polymorphic function, but it seems ok for the GraphQL API.

I will try it, thanks for the suggestion!

amiskov: I did some refactoring for the queries API used in this issue. I will try to explain how I was thinking...

## Autocomplete for the Payers Selector
I realized why I decided that a list of transfers is not a match for this query we discussed above. The thing is that I need here the list of _distinct_ profiles of those, who have transferred coins to the user, filtered by some letters from their account name. This query will be executed while the user is typing in the autocomplete field and the query result has to be small: just profiles with name, account, and avatar.

The resulted GraphQL query now looks like this:

```graphql
query ($input: ProfileInput!) {
  profile(input: $input) {
    getPayersByAccount(account: "a") {
      account
      name
      avatar
    }
  }
}

# variables
{
"input": {
  "account": "lucca"
}
}
```

The payload will be like this:

```json
{
  "data": {
    "profile": {
      "getPayersByAccount": [
        {
          "name": null,
          "avatar": "QmZTydfXVsdZbeSBJyofHGbhuZyEe4T2Fsr1ETSA7H2YsR",
          "account": "alisson"
        },
        {
          "name": "Anderson",
          "avatar": null,
          "account": "appbespiral5"
        },
        {
          "name": "fulano",
          "avatar": null,
          "account": "asdfbmkshsuw"
        }
      ]
    }
  }
}
```

If I will use transfers here, I will only get a list of transfers with _non-distinct_ profiles in the `from` field (e.g. many transfers from Karla) which is not convenient and large.

## Payment History in Transfers
I agree that `get_payment_history` is a good candidate for another `get_transfers` clauses. I tried different approaches and decided to build filters like this:

- `get_transfers` should have some more optional fields in its `args`.
- The one will be `direction` since transfers could be `INCOMING` and `OUTCOMING` (enum in GraphQL).
- Any transaction involves two parties and one of them is the user. We can specify the second party's account to get transactions related to both of them.
- And also we can specify the date in `yyyy-mm-dd` format to get transactions on a specific day.

The query would be like this:

```graphql
# get all incoming transactions from Karla on the 1st of May
query ($input: ProfileInput!, $onPage: Int!) {
  profile(input: $input) {
    transfers(first: $onPage, secondPartyAccount: "karla", date: "2020-05-01", direction: INCOMING) {
      edges {
        node {
          from {
            account
          }
          to {
            account
          }
          createdAt
        }
      }
    }
  }
}
// variables
{
  "input": {
    "account": "lucca"
  },
  "onPage": 10
}
```

We can drop any of the `transfers()` field parameters and it will work fine.

In code I pushed today we have:

- a small `get_payers_by_account` function to get profiles for the autocomplete;
- a big `get_transations` function for the `profile.transactions` field with an API to fetch transactions by direction, users involved, and the date. I decided to store them in a separate `Accounts.Transfers` module to keep code cleaner.

amiskov:
> still missing tests though 😁

@lucca65 Finally, tests have been added :) I’m eager to receive your feedback!

lib/cambiatus/accounts/transfers.ex
```diff
@@ -162,7 +162,7 @@ defmodule Cambiatus.Accounts.Transfers do
       DateTime.from_iso8601(Date.to_string(date) <> "T00:00:00Z")
 
     day_boundary_end =
-      DateTime.add(day_boundary_start, 24 * 60 * 60, :seconds)
+      DateTime.add(day_boundary_start, 24 * 60 * 60, :second)
```
lucca65, line 165: did you found this bug during tests?

amiskov: Yeah, I found it after running tests. It's not actually a bug, just a deprecation warning:

```
warning: deprecated time unit: :seconds.
A time unit should be :second, :millisecond, :microsecond, :nanosecond, or a positive integer
```

Time units [should be in a singular form](https://github.com/elixir-lang/elixir/issues/5236) since Elixir 1.8.1.

lucca65: oh, sweet!

lucca65:
@amiskov it looks good, let me know if you need any help deploying

PR #46, opened 2020-06-16
Use github actions and fix test suite

### LINKED ISSUE #45: Review tests
Our backend tests and structure are a little bit stale and unreliable. We need to improve the overall quality and cohesion of the app

RodrigoLeiteF:
Closes #45 

This PR introduces Github actions, which is a Github-integrated CI solution. It builds the project, runs tests and linters. It's a foundation for adding continuous deployment, which can be done when we improve our AWS infrastructure.

Check the Actions tab :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/cache@v2
+      - name: Setup cache
+        uses: actions/cache@v2
         with:
           path: |
             deps
@@ -15,7 +16,8 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-mix-
 
-      - uses: actions/setup-elixir@v1
+      - name: Setup Elixir
+        uses: actions/setup-elixir@v1
         with:
           otp-version: '22.2'
           elixir-version: '1.9.4'
@@ -44,7 +46,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: harmon758/postgresql-action@v1
+      - name: Setup postgres database
+        uses: harmon758/postgresql-action@v1
         with:
           postgresql db: 'cambiatus_test'
           postgresql user: 'postgres'
@@ -60,7 +63,8 @@ jobs:
         with:
           name: deps
 
-      - uses: actions/setup-elixir@v1
+      - name: Setup Elixir
+        uses: actions/setup-elixir@v1
         with:
           otp-version: '22.2'
           elixir-version: '1.9.4'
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,130 @@
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
lucca65, line 5: better to fix a version so this keeps working as time goes on

.github/workflows/test.yml
```diff
@@ -0,0 +1,130 @@
...
+            _build
+          key: ${{ runner.os }}-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+          restore-keys: |
+            ${{ runner.os }}-mix-
+
+      - name: Setup Elixir
+        uses: actions/setup-elixir@v1
+        with:
+          otp-version: '22.2'
+          elixir-version: '1.9.4'
```
lucca65, line 23: we don't run this version, we run this one: https://github.com/cambiatus/backend/blob/master/Dockerfile#L7

Its important to keep the same version during our build process as this is the one we know works. Its tempting to just go to the latest version, but we need to keep consistency

config/test.exs
```diff
@@ -32,7 +32,7 @@ config :cambiatus, Cambiatus.Auth.InvitationId,
 config :cambiatus, Cambiatus.Repo,
   database: "cambiatus_test",
   username: "postgres",
-  password: "",
+  password: "123123123",
```
lucca65, line 35: this will break everyones else `mix test` command

lucca65:
noticed that you didn't asked for anyone's review @RodrigoLeiteF, please make sure you ask for it so we got a nice email notification and it doesn't have to get to the slack bot to remind us!

config/test.exs
RodrigoLeiteF: This PR changes the test config to use that password as well. It's done because postgres doesn't actually support empty passwords. What it does is set up "trust" when running from localhost, skipping the need for a password. But since it's running on a docker container in this case, "trust" doesn't work and we need a password. After merging `mix test` should still work.

.github/workflows/test.yml
RodrigoLeiteF: According to the docs `ubuntu-latest` is an alias to Ubuntu 18.04, but I went ahead and made it explicit just in case :)

config/test.exs
lucca65: hmmm, it will break on my machine, my local Postgres instance isn't setup to use this password. But Ok, I see

.github/workflows/test.yml
lucca65: yeah! the thing is that `latest` will change over time! better to have explicit! Thanks!

PR #47, opened 2020-06-19
Pin ubuntu versions to 18.04

RodrigoLeiteF:
This fixes the failing job pushImage

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,7 @@ jobs:
           name: deps
           path: deps
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       - name: Lint project
         run: mix credo
   publishImage:
-    runs-on: ubuntu latest
```
amiskov, line 99: I wonder if the problem was in the missing hyphen here.

RodrigoLeiteF: Yep, it was. But since we decided to pin to 18.04 on the previous PR, I thought I'd go ahead and fix that as well :)

PR #50, opened 2020-07-06
Upgrade elixir to 1.9.4

### LINKED ISSUE #49: Update Elixir version to the latest
make sure our docker build process via `make build` still working properly

#### TODO
- [x] Test locally
- [ ] Update elixir version to 1.9 in staging
- [ ] Update CI config
- [ ] Update elixir version in prod

RodrigoLeiteF:
Closes #49 

1. Upgrades Elixir to 1.9.4
2. Upgrades Alpine to 3.9, as to meet the libcrypto.so.1.1 dependency introduced by the previous change

Check the Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,9 @@
 # Build Stage
 #===========
 # The version of Alpine to use for the final image
-ARG ALPINE_VERSION=3.8
+ARG ALPINE_VERSION=3.9
 
-FROM elixir:1.7.2-alpine AS builder
+FROM elixir:1.9.4-alpine AS builder
 
 # The name of your application/release (required)
 ARG APP_NAME=cambiatus
@@ -34,11 +34,11 @@ RUN mix do deps.get --only prod, deps.compile, compile
 
 RUN \
   mkdir -p /opt/built && \
-  mix release --verbose && \
-  cp _build/${MIX_ENV}/rel/${APP_NAME}/releases/${APP_VSN}/${APP_NAME}.tar.gz /opt/built && \
+  mix release && \
+  cp _build/${MIX_ENV}/prod-${APP_VSN}.tar.gz /opt/built && \
   cd /opt/built && \
-  tar -xzf ${APP_NAME}.tar.gz && \
-  rm ${APP_NAME}.tar.gz
+  tar -xzf prod-${APP_VSN}.tar.gz && \
+  rm prod-${APP_VSN}.tar.gz
 
 #================
 # Deployment Stage
@@ -51,7 +51,7 @@ ARG APP_NAME
 RUN apk update && \
     apk add --no-cache \
       bash \
-      openssl-dev
+      libssl1.1
 
 ENV REPLACE_OS_VARS=true \
     APP_NAME=${APP_NAME}
@@ -60,4 +60,6 @@ WORKDIR /opt/app
 
 COPY --from=builder /opt/built .
 
-CMD trap 'exit' INT; /opt/app/bin/${APP_NAME} foreground
+RUN mv /opt/app/bin/prod /opt/app/bin/${APP_NAME}
+
+CMD trap 'exit' INT; /opt/app/bin/${APP_NAME} start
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,9 +2,9 @@
 # Build Stage
 #===========
 # The version of Alpine to use for the final image
-ARG ALPINE_VERSION=3.8
+ARG ALPINE_VERSION=3.9
 
-FROM elixir:1.7.2-alpine AS builder
+FROM elixir:1.9.4-alpine AS builder
```
amiskov, line 7: Why did you decide to choose this version instead of the [latest](https://hub.docker.com/_/elixir) which is 1.10.3 for the moment? Do we have any issues with 1.10.X?

Dockerfile
```diff
@@ -51,7 +51,7 @@ ARG APP_NAME
 RUN apk update && \
     apk add --no-cache \
       bash \
-      openssl-dev
+      libssl1.1
```
amiskov, line 54: Can you, please, explain the change of the SSL library?

Dockerfile
lucca65: note here. Thats not something I've added explicitly in the original docker file, but we should fix OTP versions as well

Dockerfile
```diff
@@ -34,11 +34,11 @@ RUN mix do deps.get --only prod, deps.compile, compile
 
 RUN \
   mkdir -p /opt/built && \
-  mix release --verbose && \
-  cp _build/${MIX_ENV}/rel/${APP_NAME}/releases/${APP_VSN}/${APP_NAME}.tar.gz /opt/built && \
+  mix release && \
+  cp _build/${MIX_ENV}/prod-${APP_VSN}.tar.gz /opt/built && \
```
lucca65, line 38: why remove the verbose and add `prod-`

Dockerfile
RodrigoLeiteF: After upgrading Elixir I found that the application would require a very specific version of libcrypto (specifically libcrypto1.1), and after some research the only package I found that provides that module is libssl1.1 (on Alpine 3.9)

Dockerfile
RodrigoLeiteF: Good catch, I got confused by Dockerhub's UI :p

RodrigoLeiteF: Good catch, I got confused by Dockerhub's UI

Dockerfile
RodrigoLeiteF: The verbose flag seems to have been deprecated. Changing the image to 1.9.4 would cause an error wrt to it, and the output seems to be the same without it.

Adding the prod prefix is because of the `prod` release added on `mix.exs`. I couldn't find information on specifying the name of the tar file generated by mix, so I decided to use the default and rename the binary to what we have currently during the deployment phase

RodrigoLeiteF:
@lucca65 there doesn't seem to be any official Elixir 1.9 images with a fixed OTP version, and the latest one that's listed on dockerhub is 1.8.2-otp-22. Should we use 1.8.2 or pin the OTP version some other way?

lucca65:
Hmmm I took a few moments to checkout features from 1.9 and this one is actually pretty relevant

> Self-contained. A release does not require the source code to be included in your production artifacts. All of the code is precompiled and packaged. Releases do not even require Erlang or Elixir in your servers, as they include the Erlang VM and its runtime by default. Furthermore, both Erlang and Elixir standard libraries are stripped to bring only the parts you are actually using.

When they release a fixed tag for 1.9 with a specific OTP version we can change it! thanks for checking the details out @RodrigoLeiteF

Dockerfile
lucca65: ok! thanks @RodrigoLeiteF 

does this build runs on production? (mainly making sure login with invitation works, that where it usually fails)

RodrigoLeiteF: I haven't personally tested that but our test runner on Github uses Elixir 1.9.2 and all tests seem to have passed. Is that enough?

RodrigoLeiteF:
After some deliberation I've decided to pause this PR and focus on the IPFS -> S3 migration. The upgrade will take more time than we all anticipated, and while it's important, it's something we can postpone for more important features at the moment. I'll close the PR and move it back to our to-do.

PR #51, opened 2020-07-13
Deprecate IPFS endpoint for S3

### LINKED ISSUE #23: Move away from IPFS
It is very frustrating to use images in the platform. Images take a while to load, sometimes they don’t appear even if the request of saving it worked.

**Make sure to migrate shop images in MUDA**

Just use S3 

none

RodrigoLeiteF:
Closes #23 

Deprecates the IPFS endpoint in favor of an endpoint that uploads to S3

A few notes about this PR:

- This introduces an identity necessity in the form of credentials for an S3 user that can upload files to the "cambiatus-uploads" S3 bucket. I have created the "bespiral-s3" user for this purpose, but we need to discuss how to handle it's credentials when deploying.

- It also changes the format of the response. Instead of returning a hash, the endpoint returns the full URL of the file that was uploaded. This is done to prevent having to change the frontend in the future in case we need to change the upload flow.

- There are also a couple security concerns. Should we check the filesize and mime-type of the files being uploaded? In the current implementation none of these checks are done, which makes the system pretty fragile. If so, what should be the restrictions?


lucca65:
hey @RodrigoLeiteF 

here is some insights about the points you've made

- the credentials should be environment variables that we fill in the production nodes
- We should check mime types, file size, and file name, and giving uuids (or something similar) to the files we add to the bucket (removing the original naming)

RodrigoLeiteF:
> the credentials should be environment variables that we fill in the production nodes
Gotcha. The library that interacts with AWS already does this by default using `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. Do I just update the docker-compose file on the infra repository with those values?

> We should check mime types, file size, and file name, and giving uuids (or something similar) to the files we add to the bucket (removing the original naming)
👍

lucca65:
no need to configure anything on the compose files

mix.exs
```diff
@@ -51,6 +51,12 @@ defmodule Cambiatus.Mixfile do
       {:eosrpc, "~> 0.5.0"},
       {:sentry, "~> 6.4"},
       {:hashids, "~> 2.0"},
+      {:ex_aws, "~> 2.1"},
+      {:ex_aws_s3, "~> 2.0"},
+      {:poison, "~> 3.0"},
+      {:hackney, github: "benoitc/hackney", override: true},
```
amiskov, line 57: I didn't get why do we need `poison` and `hackney`... Where are those libs are used?

RodrigoLeiteF: They're dependencies for ex_aws and ex_aws_s3. Since they allow any json (posion) and http (hackney) libraries to be used, we have to add those separately

lucca65: I'm hackney is actually more performant than other libraries, but can you test if it still works with the original library? forks are dangerous

lib/cambiatus_web/controllers/s3_controller.ex
```diff
@@ -0,0 +1,67 @@
...
+
+  defp validate_file_size(file_path) do
+    info = File.lstat!(file_path)
+
+    # 2 megabytes
+    if info.size > 2_097_152 do
+      {:error, "File exceeds 2 megabytes"}
+    end
+
+    {:ok}
```
lucca65, line 12: usually on elixir when we return tuples is to keep the pattern of `{term, bit string}`. When we only want to return `:ok` we can do it without wrapping it on a tuple

lib/cambiatus_web/controllers/s3_controller.ex
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  defp validate_file_type(file_path) do
+    contents = File.read!(file_path)
+
+    case MagicNumber.detect(contents) do
+      {:ok, {:image, _}} ->
+        {:ok}
```
lucca65, line 20: for example, here its ok, but you should return a string, to keep the function return type consistent. 

Elixir is a dynamic language so it will not complain, but as you know its best to have consistency of typing

lib/cambiatus_web/controllers/s3_controller.ex
```diff
@@ -0,0 +1,67 @@
...
+    image_binary = File.read!(file_path)
+
+    operation = ExAws.S3.put_object(bucket_name, "/#{unique_filename}", image_binary)
+
+    case ExAws.request(operation) do
+      {:ok, _} ->
+        {:ok, "https://#{bucket_name}.s3.amazonaws.com/#{bucket_name}/#{unique_filename}"}
+
+      _ ->
+        {:error}
```
lucca65, line 41: same as other comments

mix.exs
RodrigoLeiteF: Do we already have an HTTP library? The reason I'm using hackney from github directly is because of a bug that was fixed on master but not released yet. Once it's released we can go back and pin the version

lib/cambiatus_web/controllers/s3_controller.ex
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  def save(conn, params) do
```
lucca65, line 45: save being here is 100% OK, as it deals with putting statuses, but all the other functions are not related to the "web" part and should be moved to another context. 

@RodrigoLeiteF check out other controllers, you will noticed that they usually are very small and with little to no business rules. That's because we follow Phoenix architecture to organize our projects. Here is a little video explaining contexts.

https://www.youtube.com/watch?v=l3VgbSgo71E

https://www.youtube.com/watch?v=D2W9gUvlaeU

RodrigoLeiteF:
I couldn't find anything about using Cowboy as an HTTP client, including on it's documentation. I guess we should stick with Hackney for now.

lib/cambiatus/upload.ex
```diff
@@ -0,0 +1,44 @@
...
+  def validate_file_size(file_path) do
+    info = File.lstat!(file_path)
+
+    # 2 megabytes
+    if info.size > 2_097_152 do
+      {:error, "File exceeds 2 megabytes"}
+    end
+
+    :ok
+  end
```
lucca65, line 31: as an exercise, can you please add the `@spec` type signature?

lib/cambiatus/upload.ex
```diff
@@ -1,6 +1,10 @@
 defmodule Cambiatus.Upload do
   @moduledoc "Handles file uploading"
 
+  @doc """
+  Saves a file on the configured Amazon S3 bucket
+  """
+  @spec save_file(String.t(), String.t()) :: {:ok, String.t()} | :error
```
lucca65, line 7: Good! Typing functions shows important stuff! You have two different types on this function. In Elm, you can't even do this, you would have to introduce a type that can have two different values. You'll see that in elixir we usually have {:ok, ""} and {:error, ""} as the most common response types. This way you will not hide the error, and give a chance for the consumer of this function to do different things depending on the error.

You did this pattern of having two different types in a lot of functions on this module, please let me know your thoughts on this, maybe I'm not seeing something here!

RodrigoLeiteF: That's fair enough. In cases where there's no relevant information to add on the success type, do we just return `{:ok, ""}` with an empty string?

lucca65: No, usually I do return something that is not an empty string. In the case of the functions you did, you could just return boolean values since the error messages are always the same. I'll comment on them.

lib/cambiatus_web/controllers/s3_controller.ex
```diff
@@ -0,0 +1,28 @@
+defmodule CambiatusWeb.S3Controller do
+  use CambiatusWeb, :controller
+
+  alias Cambiatus.Upload
+
+  def save(conn, params) do
+    with %{path: file_path, filename: filename} <- Map.get(params, "file"),
+         :ok <- Upload.validate_file_size(file_path),
+         :ok <- Upload.validate_file_type(file_path),
+         {:ok, url} <- Upload.save_file(file_path, filename) do
```
lucca65, line 10: This whole lot could be just inside the `Upload.save_file/2` function. With a return of `{:ok, String.t()} | {:error, String.t()` with the error string being the reason why it failed

lib/cambiatus/upload.ex
```diff
@@ -0,0 +1,56 @@
...
+  def validate_file_size(file_path) do
+    info = File.lstat!(file_path)
+
+    # 2 megabytes
+    if info.size > 2_097_152 do
+      {:error, "File exceeds 2 megabytes"}
+    end
+
+    :ok
+  end
```
lucca65, line 39: this could only return a bool, can be called `is_file_too_big`

lib/cambiatus/upload.ex
```diff
@@ -0,0 +1,56 @@
...
+
+  @doc """
+  Verifies if the file in the given path is an image by checking it's magic number
+  """
+  @spec validate_file_type(String.t()) :: :ok | {:error, String.t()}
+  def validate_file_type(file_path) do
```
lucca65, line 45: this could return a bool and be called something like `has_valid_mime_type`

lib/cambiatus/upload.ex
```diff
@@ -0,0 +1,56 @@
...
+    operation = ExAws.S3.put_object(bucket_name, "/#{unique_filename}", image_binary)
+
+    case ExAws.request(operation) do
+      {:ok, _} ->
+        {:ok, "https://#{bucket_name}.s3.amazonaws.com/#{bucket_name}/#{unique_filename}"}
+
+      _ ->
+        :error
+    end
+  end
```
lucca65, line 24: Here you could have something like:

```elixir
@spec save(String.t()) :: {:ok, String.t()} | {:error, String.t()}
save(file_path) do
  if is_file_too_big(file_path) do
    {:error, "File exceeds"}
  else if has_valid_mime_type(file_path)
    {:error, "Invalid mime"
  else
   save_file(file_path, Ecto.UUID.generate())
end
```

- We shouldn't allow file names, as this can contain malicious strings -- rather we can just generate an UUID by using `Ecto.UUID.generate` https://hexdocs.pm/ecto/Ecto.UUID.html#generate/0

lib/cambiatus/upload.ex
```diff
@@ -0,0 +1,56 @@
...
+    image_binary = File.read!(file_path)
+
+    operation = ExAws.S3.put_object(bucket_name, "/#{unique_filename}", image_binary)
+
+    case ExAws.request(operation) do
+      {:ok, _} ->
+        {:ok, "https://#{bucket_name}.s3.amazonaws.com/#{bucket_name}/#{unique_filename}"}
+
+      _ ->
+        :error
```
lucca65: Here I would grab the error and return it to the consumer:

```elixir
case ... do
  {:ok, _} ->
    # Return thing.
  
  {:error, reason} ->
    {:error, reason}
end
```

test/support/mocks/exaws_mock.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Cambiatus.ExAwsMock do
```
lucca65, line 1: the name of the file should follow the camel case: `ex_aws_mock.ex`

lib/cambiatus_web/controllers/upload_controller.ex
```diff
@@ -0,0 +1,28 @@
+defmodule CambiatusWeb.UploadController do
+  use CambiatusWeb, :controller
+
+  alias Cambiatus.Upload
+
+  def save(conn, params) do
+    with %{path: file_path} <- Map.get(params, "file"),
+         file_info <- File.lstat!(file_path),
+         file_contents <- File.read!(file_path),
+         {:ok, url} <- Upload.save(file_info, file_contents) do
```
lucca65, line 10: can't we just move this part to inside `Upload.save/2`?

```elixir
   file_info <- File.lstat!(file_path),
          file_contents <- File.read!(file_path),
```

RodrigoLeiteF: I moved it here to make testing the module easier. It also means we can, in the future, pass along things that aren't necessarily files in the file system to be uploaded (for example, an array buffer coming from a request or some third party service)

lib/cambiatus_web/controllers/ipfs_controller.ex
```diff
@@ -2,25 +2,8 @@ defmodule CambiatusWeb.IPFSController do
...
-        |> json(%{error: "File is required"})
-
-      _ ->
-        conn
-        |> put_status(500)
-        |> json(%{error: "Something went wrong while uploading your image"})
-    end
+    conn
+    |> put_status(:moved_permanently)
+    |> redirect(to: CambiatusWeb.UploadController.save(conn, params))
```
lucca65, line 7: Nice!!!!

lib/cambiatus_web/controllers/upload_controller.ex
lucca65: ok!

mix.exs
```diff
@@ -51,6 +51,13 @@ defmodule Cambiatus.Mixfile do
       {:eosrpc, "~> 0.5.0"},
       {:sentry, "~> 6.4"},
       {:hashids, "~> 2.0"},
+      {:ex_aws, "~> 2.1"},
+      {:ex_aws_s3, "~> 2.0"},
+      {:poison, "~> 3.0"},
+      {:hackney,
+       github: "benoitc/hackney", override: true, ref: "d8a0d979b9bdb916fe090bf1d5b076e35c2efc33"},
```
lucca65, line 58: nice

PR #53, opened 2020-07-21
Update EOSRPC and add configuration for symbol

lucca65:
It updates EOSRPC version, and configures the EOS default symbol. Its used by our action creation flow to buy resources to a newly created account. New infra uses `SYS` instead of `EOS` and this update allows us to configure which one we want

Connect to a blockchain that uses another symbol, it can be `SYS` or anything else. Configure the env var and test to see if it can create accounts properly


lucca65:
Just an update on this so won't look like it is stale. We are still validating the new infra and this will be a requirement. Until the new infra is finally the new production, this PR will remain a draft

PR #62, opened 2020-08-04
Add basic kyc and address migration

lucca65:
## What issue does this PR close
Closes #55
Closes #56

## Changes Proposed ( a list of new changes introduced by this PR)
Adds types and tables for all KYC informations that Cambiatus is introducing

## How to test ( a list of instructions on how to test this PR)
run `mix test`

# Tasks
- [x] Create migrations
- [x] Create Ecto Models
- [x] Add Costa Rica Data to our databases
- [x] Add country dependant validation structure
- [x] Add Costa Rica validation
- [x] Create GraphQL types

Issue #61, opened 2020-07-27
Add Authentication to the API


lucca65:
This issue must hold a discussion on how we should approach this. This will be necessary but it must be done right, after the blockchain receives a signed transaction it must create a new token to be used by the frontend. All logins now will need to include that new transaction, and the backend will need to check if the request made is actually valid

PR #64, opened 2020-08-10
Add signup mutation

### LINKED ISSUE #57: Add mutation for creating new users 25
We currently have a Rest API for sign_up new users.  We'll need to create a new mutation to be used on the new sign_up process. That process will need to have the same functionality we have today (Invitation support and blockchain's netlink call) and also will need to have support for all the information related to KYC -- if the community enables it -- validate it and store it on our servers

Please be aware that Costa Rica has 3 different types of physical person and 3 different types of companies, we will need to find out the data format and validate everything

RodrigoLeiteF:
Closes #57 

Introduces a new mutation for creating new users. This is necessary for the future implementation of know-your-customer functionality.

1. Open up the playground (http://localhost:4000/api/graphiql)
2. Run the mutation `createUser`

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -17,7 +17,7 @@ config :cambiatus, Cambiatus.Eos,
 config :cambiatus, Cambiatus.Repo,
   database: "cambiatus_dev",
   username: "postgres",
-  password: "",
+  password: "123123123",
   hostname: "localhost",
   port: "5432"
 
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -17,7 +17,7 @@ config :cambiatus, Cambiatus.Eos,
 config :cambiatus, Cambiatus.Repo,
   database: "cambiatus_dev",
   username: "postgres",
-  password: "123123123",
+  password: "",
   hostname: "localhost",
   port: "5432"
 
--- a/lib/cambiatus/eos.ex
+++ b/lib/cambiatus/eos.ex
@@ -35,7 +35,7 @@ defmodule Cambiatus.Eos do
         "ownerKey" => owner_key,
         "activeKey" => active_key
       }) do
-    case EOSRPC.Chain.get_account(account_name) do
+    case @eosrpc_helper.get_account(account_name) do
       {:ok, _} ->
         {:error, "Account already exists"}
 
--- a/lib/cambiatus_web/resolvers/accounts.ex
+++ b/lib/cambiatus_web/resolvers/accounts.ex
@@ -33,6 +33,16 @@ defmodule CambiatusWeb.Resolvers.Accounts do
   end
 
   @doc """
+  Updates an a user account profile info
+  """
+  @spec create_user(map(), map(), map()) :: {:ok, User.t()} | {:error, term()}
+  def create_user(_, %{input: params}, _) do
+    params
+    |> Map.new(fn {k, v} -> {Atom.to_string(k), v} end)
+    |> Cambiatus.Auth.sign_up()
+  end
+
+  @doc """
   Collects transfers belonging to the given user according various criteria, provided in `args`.
   """
   @spec get_transfers(map(), map(), map()) :: {:ok, list(map())} | {:error, String.t()}
--- a/lib/cambiatus_web/schema/account_types.ex
+++ b/lib/cambiatus_web/schema/account_types.ex
@@ -38,6 +38,7 @@ defmodule CambiatusWeb.Schema.AccountTypes do
     field(:account, non_null(:string))
     field(:email, non_null(:string))
     field(:invitation_id, non_null(:string))
+    field(:public_key, non_null(:string))
   end
 
   @desc "An input object for updating a user Profile"
--- a/lib/cambiatus_web/schema/account_types.ex
+++ b/lib/cambiatus_web/schema/account_types.ex
@@ -34,10 +34,10 @@ defmodule CambiatusWeb.Schema.AccountTypes do
 
   @desc "Input object for creating a new user account"
   input_object :create_user_input do
-    field(:name, :string)
+    field(:name, non_null(:string))
     field(:account, non_null(:string))
-    field(:email, :string)
-    field(:invitation_id, :string)
+    field(:email, non_null(:string))
+    field(:invitation_id, non_null(:string))
   end
 
   @desc "An input object for updating a user Profile"
--- a/lib/cambiatus_web/schema/account_types.ex
+++ b/lib/cambiatus_web/schema/account_types.ex
@@ -24,6 +24,20 @@ defmodule CambiatusWeb.Schema.AccountTypes do
       arg(:input, non_null(:profile_update_input))
       resolve(&Accounts.update_profile/3)
     end
+
+    @desc "Creates a new user account"
+    field :create_user, :profile do
+      arg(:input, non_null(:account_create_input))
+      resolve(&Accounts.create_user/3)
+    end
+  end
+
+  @desc "Input object for creating a new user account"
+  input_object :account_create_input do
+    field(:name, non_null(:string))
+    field(:account, non_null(:string))
+    field(:email, non_null(:string))
+    field(:invitation_id, non_null(:string))
   end
 
   @desc "An input object for updating a user Profile"
@@ -44,8 +58,8 @@ defmodule CambiatusWeb.Schema.AccountTypes do
 
   @desc "The direction of the transfer"
   enum :transfer_direction do
-    value :incoming, description: "User's incoming transfers."
-    value :outgoing, description: "User's outgoing transfers."
+    value(:incoming, description: "User's incoming transfers.")
+    value(:outgoing, description: "User's outgoing transfers.")
   end
 
   @desc "A users profile on the system"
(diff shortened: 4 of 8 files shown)


config/dev.exs
```diff
@@ -17,7 +17,7 @@ config :cambiatus, Cambiatus.Eos,
 config :cambiatus, Cambiatus.Repo,
   database: "cambiatus_dev",
   username: "postgres",
-  password: "",
+  password: "123123123",
```
lucca65, line 20: you can just change your local Postgres database here to have no password, I've done this on my setup

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -24,6 +24,20 @@ defmodule CambiatusWeb.Schema.AccountTypes do
...
+
+    @desc "Creates a new user account"
+    field :create_user, :profile do
+      arg(:input, non_null(:account_create_input))
+      resolve(&Accounts.create_user/3)
+    end
+  end
+
+  @desc "Input object for creating a new user account"
+  input_object :account_create_input do
```
lucca65, line 36: since the name of the mutation is `create_user` we should keep a consistent name for the init as well

lib/cambiatus_web/resolvers/accounts.ex
```diff
@@ -32,6 +32,16 @@ defmodule CambiatusWeb.Resolvers.Accounts do
     end
   end
 
+  @doc """
+  Updates an a user account profile info
+  """
+  @spec create_user(map(), map(), map()) :: {:ok, User.t()} | {:error, term()}
+  def create_user(_, %{input: params}, _) do
+    params
+    |> Map.new(fn {k, v} -> {Atom.to_string(k), v} end)
```
lucca65, line 41: this is interesting. I've seen atomifications using different methods, but I like this one!

test/cambiatus_web/schema/resolvers/accounts_test.exs
```diff
@@ -10,6 +10,65 @@ defmodule CambiatusWeb.Schema.Resolvers.AccountsTest do
...
+        }
+      }
+
+      query = """
+      mutation($input: AccountCreateInput!){
+        createUser(input: $input) {
+          account
+          email
+          name
+        }
```
lucca65, line 34: gotta love GraphQL!

test/cambiatus_web/schema/resolvers/accounts_test.exs
```diff
@@ -10,6 +10,65 @@ defmodule CambiatusWeb.Schema.Resolvers.AccountsTest do
...
+      res = conn |> post("/api/graph", query: query, variables: variables)
+
+      response = json_response(res, 200)
+
+      assert response["data"]["createUser"]["account"] == variables["input"]["account"]
+      assert response["data"]["createUser"]["email"] == variables["input"]["email"]
+      assert response["data"]["createUser"]["name"] == variables["input"]["name"]
+    end
+
+    test "creates a user with minimal parameters", %{conn: conn} do
```
lucca65, line 47: This test got my attention. Currently we have two different API endpoints to do our signup process. We only allow signup for existing Eos accounts, and the frontend make sure we first create an account an later it does the sign up. This may be a good opportunity to merge both. Any thoughts? 

@amiskov I would love your input on this. How complicated it would be on the frontend?

I bring this now because I can't think of any scenario where we can create EOS accounts without creating an Cambiatus user.

With the new flow we would only have:

Sign in (user already have an Eos account) and Sign on (user don't have an Eos account)

lucca65: We could even check to see if the user already have an EOS account and create it if needed. The error report would become simpler too (we would skip the HTTP call the frontend does in order to know if the account name was already taken).

amiskov: > Currently we have two different API endpoints to do our signup process.

Do you mean Rodrigo's implementation via GraphQL and this one?

```elixir
post("/auth/sign_up", AuthController, :sign_up)
```

> This may be a good opportunity to merge both.

Does that mean using one GraphQl endpoint?

@lucca65 If I understand it right, your suggestion sounds reasonable and looks simpler.

Now we check if the EOS account exists on the JS part. But if we will use GraphQL endpoint for doing the sign-up process, then we can use Elm and it will be more consistent and reliable. And all the logic will be on the backend, which is better.

lucca65: > Do you mean Rodrigo's implementation via GraphQL and this one?

Yes!

> Does that mean using one GraphQl endpoint?

yes also!

I was thinking like this because it will simplify our frontend and will allow us to do #61 much easier too, since we'll be able to deliver a auth token directly on a single api call.

Allow us too to control changes easily too

config/dev.exs
amiskov: Or we can start using ENV variables here or ignored files with related keys, like [12 factor suggests](https://12factor.net/config). Because there will be always differences in our dev configs :)

test/cambiatus_web/schema/resolvers/accounts_test.exs
```diff
@@ -371,9 +439,9 @@ defmodule CambiatusWeb.Schema.Resolvers.AccountsTest do
 
       %{
         "data" => %{
-           "profile" => %{
-              "getPayersByAccount" => payers
-           }
+          "profile" => %{
+            "getPayersByAccount" => payers
+          }
         }
```
amiskov, line 445: @RodrigoLeiteF did you use some kind of autoformatter on this file? :) Code looks nicer here and there :)

test/cambiatus_web/schema/resolvers/accounts_test.exs
amiskov: Sounds like we should definitely do that :)

config/dev.exs
RodrigoLeiteF: Right, this was an accidental commit. I'll just reverse it for now and we can look into it some other time. Sorry :p

test/cambiatus_web/schema/resolvers/accounts_test.exs
RodrigoLeiteF: Yep, my editor is configured to run an auto formatted on save. It's very handy

amiskov: @RodrigoLeiteF Cool! Which one do you use? I guess it's not dependent on Emacs?
I'm using JetBrains' WebStorm. It's incredible for Elm and JS, but pretty poor for Elixir...

RodrigoLeiteF: I believe it just runs `mix format`. If WebStorm supports custom on-save hooks, you can just tell it to run that on the saved file and hopefully it should work

test/cambiatus_web/schema/resolvers/accounts_test.exs
RodrigoLeiteF: I had a talk with Lucca on Slack about this, and this is the proposed change:

### Current flow

1. User wants to create an account
2. User fills the form and submits
3. Frontend validates the information with the blockchain
4. If valid, frontend creates the account on the blockchain
5. Frontend sends request to backend to create a user
6. Backend creates a user, associates it with a public key and a community

### Proposed flow
1. User wants to create an account
2. User fills the form and submits
3. Frontend sends the information to backend
5. Backend validates the information, creates an account on the blockchain and a user on the database
6. Backend associates all the information and returns the user information (in the future, also returns an authentication token

I believe this is a good change, as it streamlines the process and leaves the frontend completely naive to the signup process.

@lucca65 Does this sound correct?

lucca65: that's it @RodrigoLeiteF!

I think this will be a good change for future API growth and change

lib/cambiatus/eos.ex
```diff
@@ -35,7 +35,7 @@ defmodule Cambiatus.Eos do
         "ownerKey" => owner_key,
         "activeKey" => active_key
       }) do
-    case EOSRPC.Chain.get_account(account_name) do
+    case @eosrpc_helper.get_account(account_name) do
```
lucca65, line 38: @RodrigoLeiteF I've noticed it just now that this function don't exist. Not sure why you changed this to Helper instead of chain, can you please explain? I know it was already merged, but I can fix this on #62

RodrigoLeiteF: It does, I added it to the helper https://github.com/cambiatus/backend/pull/64/files#diff-f0ff3b87734d405e7c288a15777b2e22

I did it so we can test the function properly. Otherwise, it would always error since tests aren't run in an EOS context.

lucca65: but that function is not on helper on the eosrpc library, so it may pass the tests but it will not succeed during execution

RodrigoLeiteF: Oh, I see. I can submit a fix PR if you'd prefer

lucca65: I've just deployed a version using this PR and the app breaks. I'll @ you at the sentry notification. 

@RodrigoLeiteF please be more attentive with the tests. Remember that you have access to the app.cambiatus.io env, and you can build docker images from your development version and check if its working. Looks like that you didn't tested it.

Please, improve the qualities of your deliveries, we count on you do deliver stuff that works, I understand if you need time or if you want assistance, but test what you do

PR #67, opened 2020-08-19
Fix mocking of EOSRPC.Chain

RodrigoLeiteF:
Closes N/A

Mocks the EOSRPC.Chain module so the create_account endpoint can be properly tested

N/A


test/support/mocks/eosrpc_chain_mock.ex
```diff
@@ -0,0 +1,8 @@
+defmodule EOSRPC.ChainMock do
+  @moduledoc "Mocked implementation of EOSRPC Chain"
+  @behaviour EOSRPC.Chain
```
lucca65, line 3: this behaviour don't exist in the version the project is configured (`0.6.0`). You need to add the callbacks there first

lucca65:
@RodrigoLeiteF the problem was the alpine version we used, after a lot of digging and testing, this works on both development and prod. After all it was my bad, updating elixir version without testing it on production. and sorry for all the mess. the mock was still wrong and not working on prod, so I guess the testing thing goes to us both ):

PR #69, opened 2020-08-24
Fix/invitations

lucca65:
N/A

- [x] Add index to guarantee uniqueness of the network table (community and account)
- [x] make `mix.exs` project version match the docker version
- [x] Add signup with and without invitation on graphql
- [x] greatly improves error handling on auth functions
- [x] added account constraints validations
- [x] Adjust tests

- as always, `mix test`
- be logged in on the frontend, create a new invitation. Make sure it works and copy the link
- now log out
- open the link of the invitation you just created
- make sure you are able to create a new account and be invited correctly to the community


amiskov (changes requested):
Looks good to me!

One thing I noticed is that the test in `test/cambiatus_web/controllers/auth_controller_test.exs:22` seems to be broken:
image

Test waits for the `400 Bad Request` code but got `500 Internal Server Error`. I guess it happens because instead of [`%{account: 1}`](https://github.com/cambiatus/backend/blob/c4ff61e6b0ae47b3b72e66217d2c5defd5b146f8/test/cambiatus_web/controllers/auth_controller_test.exs#L24) for mocking the bad request we need `%{user: %{account: 1}}`. Because only this pattern is handled by the [`sign_in`](https://github.com/cambiatus/backend/blob/c4ff61e6b0ae47b3b72e66217d2c5defd5b146f8/lib/cambiatus_web/controllers/auth_controller.ex#L9) function.

lucca65:
@amiskov that a weird behaviour, this test pass on my machine and even passed on the CI... It should be failing. How are you testing those @amiskov ?

amiskov:
> How are you testing those?

@lucca65 nothing special... I just check out the branch, pull the changes, run `mix deps.clean`, then `mix deps.get` and then `mix test`. I use `asdf` and the `.tool-versions` file, so the Elixir version is the same, 1.8.2.

Running this specific test like this `mix test test/cambiatus_web/controllers/auth_controller_test.exs:22` gives me this error...

lucca65:
@amiskov I really don't know what to do here. Docker builds, our CI pass, my local tests pass, tests on docker pass. Any ideas?

amiskov:
> I really don't know what to do here. Docker builds, our CI pass, my local tests pass, tests on docker pass. Any ideas?

I think it's ok to merge this PR :) That test is not that critical.

lucca65:
ok @amiskov I've re-asked for the review

amiskov (approved):

PR #78, opened 2020-09-10
User KYC data deletion mutation

### LINKED ISSUE #58: Allow users to delete their KYC information 6
We need to add a mutation to delete KYC information from the user profile

victorgdev:
Closes #58 

KYC user data deletion mutation

On GraphiQL Playground run the following mutations

```
// KYC data mutation 
mutation {
        deleteKyc(input: 
          { account : "your_account_id" }) {
                    reason
    		    status
        }
}

// Address data mutation 
mutation {
        deleteAddress(input: 
          { account : "your_account_id" }) {
                    reason
    		    status
        }
}
```


lib/cambiatus/kyc/kyc.ex
```diff
@@ -20,4 +24,12 @@ defmodule Cambiatus.Kyc do
         {:ok, country}
     end
   end
+
+  @doc """
+  Deletes the kyc_data
+  """
+  def kyc_data_deletion(params) do
+    Repo.get_by(KycData, params.account_id)
+    Repo.delete(params)
```
amiskov, line 33: I guess there should be something like this:

```elixir
kyc = Repo.get_by(KycData, params.account_id)
Repo.delete(kyc)
```

First, we get the actual KYC record which we want to delete and then pass it to the function, which handles the actual deleting from the DB.

lib/cambiatus/kyc/kyc.ex
```diff
@@ -20,4 +24,12 @@ defmodule Cambiatus.Kyc do
         {:ok, country}
     end
   end
+
+  @doc """
+  Deletes the kyc_data
+  """
+  def kyc_data_deletion(params) do
```
amiskov, line 31: Oh, and there's a common convention to give function names starting with verbs: `delete_kyc` or something like this.

lib/cambiatus_web/schema/kyc_types.ex
```diff
@@ -44,4 +44,33 @@ defmodule CambiatusWeb.Schema.KycTypes do
...
+
+  @desc "Input for user's KYC data deletion"
+  input_object :kyc_data_deletion do
+    field(:account_id, non_null(:string))
+    field(:country_id, non_null(:string))
+    field(:user_type, non_null(:string))
+    field(:document_type, non_null(:string))
+    field(:document, non_null(:string))
+    field(:phone, non_null(:string))
+  end
```
lucca65, line 66: you can just send the account here

lucca65 (changes requested):

lib/cambiatus_web/schema/kyc_types.ex
amiskov: Yep, that's enough for us to have `account_id` to find the KYC record which we want to delete:
```elixir
  input_object :kyc_data_deletion do
    field(:account_id, non_null(:string))
  end
```

The actual mutation could look like this:

```graphql
mutation ($input: KycDataDeletion!) {
  deleteKyc(input: $input) {
    document
    userType
    # some info from the deleted KYC record
    # that we may need on the frontend
  }
}

# Input
{
  "input": {
    "accountId": "andreymiskov"
  }
}
```

lib/cambiatus_web/schema/kyc_types.ex
```diff
@@ -80,6 +80,17 @@ defmodule CambiatusWeb.Schema.KycTypes do
...
 
+  @desc "Input for deleting KYC fields"
+  input_object :kyc_deletion_input do
+    field(:account_id, non_null(:string))
+  end
+
+  @desc "Input for deleting address fields"
+  input_object :address_deletion_input do
+    field(:account_id, non_null(:string))
+  end
```
amiskov, line 91: Since these inputs are the same, I think it's ok to use only `:kyc_deletion_input` for both KYC and Address deletion.

lib/cambiatus/kyc/kyc.ex
```diff
@@ -66,4 +71,20 @@ defmodule Cambiatus.Kyc do
       {:error, %{errors: errors_list}} -> {:error, "#{inspect(errors_list)}"}
     end
   end
+
+  @doc """
+  Deletes the kyc_data
+  """
+  def delete_kyc(params) do
+    kyc = Repo.get_by(KycData, params.account_id)
```
amiskov, line 79: Unfortunately, this line won't work. Here's a little explanation.

After the evaluation, `params.account_id` becomes a string like `"victorgoncal"`. But we have to explain what `"victorgoncal"` is to the function `get_by`. So instead of a string, it [should be a map](https://hexdocs.pm/ecto/Ecto.Repo.html#c:get_by/3) like this: `%{account_id: "victorgoncal"}`.

If we check what is stored in `params` then we will see that it's lust this map, so this will work fine:

```elixir
kyc = Repo.get_by(KycData, params)
```

BTW, you can use [`IO.inspect(params)`](https://hexdocs.pm/elixir/IO.html#inspect/2) to check what is stored in variables. The output will be visible in the terminal where the Phoenix is running.

test/cambiatus_web/schema/resolvers/kyc_test.exs
```diff
@@ -140,5 +140,89 @@ defmodule CambiatusWeb.Schema.Resolvers.KycTest do
       assert updated_kyc["document_type"] == new_kyc.document_type
       assert updated_kyc["user_type"] == new_kyc.user_type
     end
+
+    test "deletes kyc for the given account", %{conn: conn} do
+      kyc = insert(:kyc_data)
+      user = kyc.account
+
+      new_kyc = delete(:kyc_data, %Plug.Conn{})
```
amiskov, line 148: This line is unnecessary. The GraphQl mutation itself will delete the KYC record.

test/cambiatus_web/schema/resolvers/kyc_test.exs
```diff
@@ -140,5 +140,89 @@ defmodule CambiatusWeb.Schema.Resolvers.KycTest do
...
+
+      variables = %{
+        "input" => %{
+          "account_id" => user.account
+        }
+      }
+
+      query = """
+      mutation ($input: KycDeleteInput!) {
```
amiskov, line 157: Actually, your input object has the name `KycDeletionInput`(the `:kyc_deletion_input` symbol from `lib/cambiatus_web/schema/kyc_types.ex`).

test/cambiatus_web/schema/resolvers/kyc_test.exs
```diff
@@ -140,5 +140,89 @@ defmodule CambiatusWeb.Schema.Resolvers.KycTest do
...
+        deleteKyc(input: $inputKyc) {
+          user_type
+          document
+          document_type
+          phone
+          country_id
```
amiskov, line 163: The `:kyc_data` object had `country` field, not `country_id`. And the country itself is also an object with fields, so this is the right way to get the country id: `country { id }`.

test/cambiatus_web/schema/resolvers/kyc_test.exs
```diff
@@ -140,5 +140,89 @@ defmodule CambiatusWeb.Schema.Resolvers.KycTest do
...
+
+      res = conn |> post("/api/graph", query: query, variables: variables)
+
+      %{
+        "data" => %{
+          "deleteKyc" => deleted_kyc
+        }
+      } = json_response(res, 200)
+
+      assert deleted_kyc["user_type"] == new_kyc.user_type
```
amiskov, line 177: Instead of `new_kyc` you should use just `kyc`, which you created earlier on line `145`:

```elixir
assert deleted_kyc["user_type"] == kyc.user_type
```

Here you check if your original `kyc` you created has the same fields as the record you deleted with the mutation.

test/cambiatus_web/schema/resolvers/kyc_test.exs
```diff
@@ -140,5 +140,89 @@ defmodule CambiatusWeb.Schema.Resolvers.KycTest do
...
+      assert deleted_kyc["document"] == new_kyc.document
+      assert deleted_kyc["document_type"] == new_kyc.document_type
+      assert deleted_kyc["phone"] == new_kyc.phone
+      assert deleted_kyc["country_id"] == new_kyc.country_id
+      assert deleted_kyc["is_verified"] == new_kyc.is_verified
```
amiskov, line 182: Actually, it's not necessary to check all the fields. You can check several most important ones, like `user_type`, `document_type`, and `document`.

test/cambiatus_web/schema/resolvers/kyc_test.exs
```diff
@@ -140,5 +140,89 @@ defmodule CambiatusWeb.Schema.Resolvers.KycTest do
...
+    end
+
+    test "deletes address for the given account", %{conn: conn} do
+      address = insert(:address)
+      user = address.account
+
+      new_address = delete(:address, %{account: nil})
```
amiskov, line 189: Comments for KYC test are also related to the Address test.

amiskov (changes requested):

lib/cambiatus/kyc/kyc.ex
victorgdev: 🙌 You're the man!! Tks!

amiskov (approved):
Looks good to me!

lib/cambiatus_web/schema/kyc_types.ex
```diff
@@ -80,6 +80,12 @@ defmodule CambiatusWeb.Schema.KycTypes do
     field(:zip, non_null(:string))
   end
 
+  @desc "Input for deleting KYC fields"
+  input_object :kyc_deletion_input do
+    field(:account_id, non_null(:string))
```
lucca65, line 85: `account_id` don't make a lot of sense, in some places its called only account (profile object, sign_up_input) etc. Since the account is already an ID this is redundant

lucca65 (changes requested):

lib/cambiatus_web/schema/kyc_types.ex
```diff
@@ -92,5 +98,17 @@ defmodule CambiatusWeb.Schema.KycTypes do
       arg(:input, non_null(:address_update_input))
       resolve(&Kyc.upsert_address/3)
     end
+
+    @desc "A mutation to delete user's kyc data"
+    field :delete_kyc, :kyc_data do
```
lucca65, line 103: I don't think the return type here is valid, if we return the deleted KYC it will not make no sense, we should return an status. See sign up for an example:

```elixir

  object :sign_up do
    field(:status, non_null(:sign_up_status))
    field(:reason, non_null(:string))
  end
```

https://github.com/cambiatus/backend/blob/0e7d9c82c59a78ca3790cb020705b0426fcc2115/lib/cambiatus_web/schema/account_types.ex#L72

lucca65 (changes requested):

lib/cambiatus_web/schema/kyc_types.ex
lucca65: it is still returning `:kyc_data` and not the status. Please test it first!

lucca65 (changes requested):

lucca65 (changes requested):
Nice! we are getting there. Still one problem though. If the frontend asks to delete KYC for a user that don't have it (it may happen due to having two sessions opened) the backend crashes. It also crashes if we inform an account that don't exist.

we can use the status to inform the success of failure along with a message. You can always check the sign up mutation for an example on how to do that!

I'm pushing you to make this feature complete @victorgdev! I'm sure you can nail it! game on! 💪

lib/cambiatus_web/schema/kyc_types.ex
amiskov: @victorgdev @lucca65 That's why it's always better to discuss the API before the actual implementation :)

@lucca65 how cool is that that we could have a code snippet with a line number just pasting the URL for the line of code in the GitHub comment like you did! I'll use it all the time from now on :)

lib/cambiatus_web/resolvers/kyc.ex
```diff
@@ -20,4 +24,32 @@ defmodule CambiatusWeb.Resolvers.Kyc do
...
+  @spec delete_kyc(map(), map(), map()) :: {:ok, KycData.t()} | {:error, term()}
+  def delete_kyc(_, %{input: params}, _) do
+    params
+    |> Kyc.delete_kyc()
+    |> case  do
+      {:error, reason} ->
+        Sentry.capture_message("KYC deletion failed", extra: %{error: reason})
+        {:ok, %{status: :error, reason: reason}}
+
+      _ ->
```
lucca65, line 37: maybe we should be sure it worked with another match -> `{:ok, _}`

lucca65 (changes requested):

lib/cambiatus/kyc/kyc.ex
```diff
@@ -66,4 +71,24 @@ defmodule Cambiatus.Kyc do
       {:error, %{errors: errors_list}} -> {:error, "#{inspect(errors_list)}"}
     end
   end
+
+  @doc """
+  Deletes the kyc_data
+  """
+  def delete_kyc(params) do
+    case Repo.get_by(KycData, account_id: params.account) do
+      nil -> {%KycData{is_verified: false}}
```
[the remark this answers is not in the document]
victorgdev: We'll make the changes right away

lucca65 (approved):
+1!

PR #80, opened 2020-09-14
Add id field to country, state, city and neighborhood

RodrigoLeiteF:
Closes N/A

Returns unique identifiers on address queries to be used by the frontend on mutations

1. Open the graphql playground
2. Use the `country` query

### DIFF
--- a/lib/cambiatus_web/schema/kyc_types.ex
+++ b/lib/cambiatus_web/schema/kyc_types.ex
@@ -24,18 +24,21 @@ defmodule CambiatusWeb.Schema.KycTypes do
 
   @desc "KYC supported countries"
   object :country do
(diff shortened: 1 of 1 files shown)


lucca65 (changes requested):
none of those IDs are strings

amiskov:
> none of those IDs are strings

@lucca65 @RodrigoLeiteF I also used `:string` for those `id`s in the mutations for address and KYC data. Don't remember why I decided to do so, but it doesn't seem to be right. I think we should use `non_null(:id)`. Only `field(:account_id, non_null(:string))` seems to be appropriate since account id is a twelve-letter word.

lucca65:
+1 @amiskov

lucca65 (approved):

PR #70, opened 2020-10-01
Remove the old sign up API

RodrigoLeiteF:
Instead of simply removing the route, I opted to instead redirect to our graphql endpoint using the 301 Moved Permanently status code

Closes #70

### DIFF
--- a/lib/cambiatus_web/controllers/auth_controller.ex
+++ b/lib/cambiatus_web/controllers/auth_controller.ex
@@ -19,11 +19,10 @@ defmodule CambiatusWeb.AuthController do
     end
   end
 
-  @spec sign_up(Plug.Conn.t(), map) ::
-          {:error, any} | {:ok, nil | [%{optional(atom) => any}] | map} | Plug.Conn.t()
-  def sign_up(conn, %{"user" => user_params}) do
-    with {:ok, %User{} = user} <- Auth.sign_up(user_params) do
-      render(conn, "auth.json", user: user)
-    end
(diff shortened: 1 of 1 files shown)


lucca65:
it says the issue it closes is 70, but that is the number of the PR itself

RodrigoLeiteF:
> it says the issue it closes is 70, but that is the number of the PR itself

Right. I accidentally clicked the wrong button on my Github integration software and it turned issue #70 into a PR for itself, with the same number. It won't happen again. Do I need to open a new issue?

amiskov:
> it turned issue #70 into a PR for itself, with the same number.

It looks interesting! Sometimes it may be annoying to manage both the issue and the PR. There may be helpful discussions in both. E.g. in an issue we started discussing some details before the actual implementation in code and then when there are some code written we continue the discussion in the PR, linking comments from the corresponding issue. I wonder if it may be useful for us to actually adopt converting issues to PRs as we start to write the code!

> I accidentally clicked the wrong button on my Github integration software

@RodrigoLeiteF which one do you use? Didn't see this functionality before.

RodrigoLeiteF:
> > I accidentally clicked the wrong button on my Github integration software
> 
> @RodrigoLeiteF which one do you use? Didn't see this functionality before.

Magit with Forge. With the magit window open, press `@ c u`. I don't know if it's possible to do it through the Github UI though, unfortunately

lucca65:
Not sure what your GitHub integration did @RodrigoLeiteF but I can't approve this PR for some reason 🤷

amiskov (approved):

PR #102, opened 2020-10-25
Feature/claim proof

### LINKED ISSUE #96: Update the GraphQL API to include the new photo proof fields

lucca65:
Closes #96 

- Fix migration from pr #99 
- Add GraphQL types


amiskov (approved):
Looks good!

@lucca65 can you please explain why did you decide to remove `20200519201353_claim_status_data_analysis_update.exs` migration? This is the only thing I didn't get in this PR.

lucca65:
@amiskov this is because that migration did updates in the table. This is not considered a good practice because when we add new fields to the model, the migration starts failing. The model will specify the new fields, but the database at that point won't have the fields. This generates an error.

The right way to do data migrations is to create a Mix Task, similar to what we do [here](https://github.com/cambiatus/backend/blob/29cd979c1983f9986f8cff19f5a43f260dda8b5b/priv/repo/seeds.exs)

Sadly at the time, we did it the wrong way and it only appeared when we changed the claim model, which happened in this PR

PR #103, opened 2020-10-28
Fix analysis page listing

amiskov:
Closes an issue with Claims listing on the Analysis page from the [frontend/409](https://github.com/cambiatus/frontend/issues/409)

Change the query for retrieving all claims to analyze for the given validator from the given community.

- Login as a validator for some Claim.
- Go to the `dashboard/analysis` page, try to approve/disapprove some Claims.
- The number of Claims on this page should stay the same, only the status for the current user should change.
- Other verifiers should also have the right amount of Claims on their Analysis pages.

### DIFF
--- a/lib/cambiatus/commune/commune.ex
+++ b/lib/cambiatus/commune/commune.ex
@@ -254,11 +254,8 @@ defmodule Cambiatus.Commune do
       on: o.id == a.objective_id,
       join: v in Validator,
       on: v.action_id == c.action_id,
-      left_join: ch in Check,
-      on: ch.claim_id == c.id,
       where: o.community_id == ^community_id,
       where: v.validator_id == ^account,
-      where: fragment("?.validator_id = ? OR ?.claim_id IS NULL", ch, ^account, ch),
       order_by: [desc: c.created_at]
     )
   end


lib/cambiatus/commune/commune.ex
```diff
@@ -254,11 +254,8 @@ defmodule Cambiatus.Commune do
       on: o.id == a.objective_id,
       join: v in Validator,
       on: v.action_id == c.action_id,
-      left_join: ch in Check,
-      on: ch.claim_id == c.id,
       where: o.community_id == ^community_id,
       where: v.validator_id == ^account,
-      where: fragment("?.validator_id = ? OR ?.claim_id IS NULL", ch, ^account, ch),
```
amiskov, line 261: @lucca65 you [were right](https://github.com/cambiatus/frontend/issues/409#issuecomment-714655573)! The problem indeed was in the `IS NULL` checking in this line.

The thing is that `ch.claim_id` is NULL only if the claim wasn't voted at all. But this field becomes non-NULL when _any_ validator approves/disapproves the Claim. That means that when some validator approves the Claim then it disappears from the Analysis page for all other validators.

I decided to remove `left_join` with `checks` because it looks like a redundant artifact from the past. The reason we have this join is the subquery in the `fragment`. But `ch.validator_id = 'accountname'` is already valid, because `Claim` has may `Check` in the schema. And I already wrote about `IS NULL` above.

lucca65 (approved):
sweet! our users will be glad to finally have claims working properly!!

PR #106, opened 2020-11-03
Resize uploaded images with Mogrify

### LINKED ISSUE #85: Some images won't upload from iPhones
Some photos from the iPhone photo gallery won't upload to `xes.cambiatus.io`, `app.cambiatus.io`, and `staging.cambiatus.io`. This [Slack thread](https://cambiatus.slack.com/archives/CA83HJAAD/p1601025645019500?thread_ts=1600966822.005700&cid=CA83HJAAD) has an example of a "broken" photo. _But the same photo could be uploaded form the `localhost` (`:3000` frontend, `:4000` backend)_.

While trying to upload a "broken" photo from any browser this error appears and the photo won't be saved to S3:
image

## Possible Problem
Some of the problematic photos were autocorrected (adjusted brightness, contract, etc.). Probably these photos have the metadata which breaks something.

I also suspect iCloud. On the iPhone, there’s an option to store only previews on the phone and preload originals from iCloud by the need. All problematic photos were loaded from iCloud right before uploading it to the website.

## How to Fix the Photo on the iPhone
This problem could be solved by choosing image size other than actual (iPhone has this option in the select image dialog for upload forms):
image

## How to Fix on the backend
It seems that processing/resizing images before uploading to S3 will help. I think we need to do that because now it’s pretty common for us to have a 2000px wide picture for 80x80px avatar. The same is for the Shop images. We can save much space on S3 and significantly speed up the loading of the website!

As an experiment, I tried to convert the problematic images with [Mogrify](https://hex.pm/packages/mogrify) (a wrapper around ImageMagick) in Elixir and the processed image was uploaded correctly.

amiskov:
Closes #85.

Big images will be reduced to 1200px on the biggest side. Small images won't be changed.

⚠️  Make sure you have [imagemagick](https://imagemagick.org/) installed in the system.

Try to upload some big images from your phone. The image should be uploaded and have a smaller size.


amiskov:
@lucca65 I consider this PR as a quick fix that will be replaced with a more robust solution in the nearest future.

There's a nice package called [Waffle](https://github.com/elixir-waffle/waffle) which is pretty common for uploading images to S3. It gives pretty convenient abstractions and is quite widely used in the community. I need some time to learn how to use it properly in Cambiatus so I decided to start from this PR.

lucca65 (approved):
:shipit:

PR #115, opened 2020-11-30
Recess Updates

### LINKED ISSUE #66: Add account creation validations
Add sign up validation for the account creation process, now we don't make any validations for the input values.

### LINKED ISSUE #113: Improve backend Sign Up process
- Make sure to validate the KYC data/address before netlink
- Allow KYC/Address mutation to happen in a single request with the signup

amiskov:
Closes #113. There's also the [frontend PR](https://github.com/cambiatus/frontend/pull/455) for this issue.
Closes #66 

- Update sign up mutation to handle Default, Natural, and Juridical account with optional data:
  - Default account has no KYC and no Address (no optional data);
  - Natural account has KYC and hasn't Address.
  - Juridical account has both KYC and Address.
- Improve Shop Queries 
  - Refactor and rename Shop 
- Small refactor on Profile

- Run `mix test`.
- Try to create different accounts: default, natural, and juridical. Make sure to post some error data and check that accounts with the errors won't be created.


amiskov:
@lucca65 can you please evaluate the approach I suggested here? I think it would be appropriate to have three mutations for creating default, natural+kyc, and juridical+kyc+address accounts.

The [frontend for this task](https://github.com/cambiatus/frontend/pull/455) is pretty much done. We only need to show better error messages and maybe slightly refactor some code for sending mutations.

As for me, I like it! The code on the frontend looks much cleaner. The backend is still in the early stage, the code is mostly boilerplate and debugging, but I get you'll get the idea!

lucca65 (commented):
Hey @amiskov i would do it a little differently, maybe by having only one mutation but inferring the type of mutation based on given params

amiskov:
@lucca65 I tried your suggestion and it worked really well! The code became much cleaner both on the frontend and the backend.

lucca65:
@amiskov I still have to test the whole thing and update the graphql interfaces but please take a look on the pipeline, I think it looks much better now!

amiskov (commented):
Looks very promising so far! Great work!

I have only a few minor suggestions: one for naming and a couple of additions for the comment blocks.

lib/cambiatus/auth/sign_up.ex
```diff
@@ -0,0 +1,171 @@
...
+
+  def create_eos_account({:error, _} = error), do: error
+  def create_eos_account({:error, _, _} = error), do: error
+
+  def create_eos_account(%{account: account, public_key: public_key} = params) do
+    case Eos.create_account(account, public_key) do
+      {:ok, _} ->
+        params
+
+      {:error, "Account already exists"} ->
```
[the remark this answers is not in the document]
lucca65: this will require a change in the EOS module, because this is not passing text, its actually matching text

lib/cambiatus/auth/sign_up.ex
```diff
@@ -0,0 +1,171 @@
...
+
+  @doc """
+  Validates the given params. Depending on the structure do different validations
+  """
+  def validate_all(params) do
+    params
+    |> validate(:account)
+    |> validate(:changeset)
+    |> validate(:user_type)
+    |> validate(:invite)
```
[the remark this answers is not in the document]
lucca65: ok! I'll change it

lib/cambiatus/auth/sign_up.ex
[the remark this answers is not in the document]
lucca65: let me know your thoughts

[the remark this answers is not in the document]
amiskov: > this will require a change in the EOS module because this is not passing text, its actually matching text

If I understand it right, this text is from `Cambiatus.Eos.create_account`:

https://github.com/cambiatus/backend/blob/bdbccad4c1657f8f560fc32493d39a2f92e566ae/lib/cambiatus/eos.ex#L47-L49

I decided to leave the comment because we have `"Account already exists"` passed from `Cambiatus.Eos.create_account`, matched in `Cambiatus.Auth.SignUp.create_eos_account` and at the end, it becomes an atom `:user_already_registered`:
https://github.com/cambiatus/backend/blob/172767693e223493493e83b26aee3f293df0c42c/lib/cambiatus/auth/sign_up.ex#L126-L127

It looks like we may use only the atom everywhere. How do you think?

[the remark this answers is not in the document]
lucca65: I wasn't thinking about refactoring Eos/Wallet types in this PR to be honest, it can be done surely, it will just make it bigger....

but Ok, I'll work on it! 👍

lucca65:
@amiskov PR is ready for review

lib/cambiatus/accounts/user.ex
```diff
@@ -45,7 +45,7 @@ defmodule Cambiatus.Accounts.User do
     has_one(:kyc, KycData, foreign_key: :account_id)
   end
 
-  @required_fields ~w(account)a
+  @required_fields ~w(account email name)a
```
lucca65, line 48: This requires some changing on the DataCase (old factory we use for some test cases) but it will make validation easier during SignUp

lib/cambiatus/auth/sign_up.ex
```diff
@@ -0,0 +1,213 @@
...
+      {:error, _} = error ->
+        error
+
+      {:error, _, _} = error ->
+        error
+
+      _result ->
+        {:ok, Accounts.get_user(params.account)}
+    end
+  end
```
lucca65, line 38: This pipeline and the one used on `validate_all/1` is basically a way to only do validation if previous steps worked.

lib/cambiatus/auth/sign_up.ex
```diff
@@ -0,0 +1,213 @@
...
+      {:ok, _} -> params
+      _ -> {:error, :failed_to_add_kyc}
+    end
+  end
+
+  def create_kyc(%{kyc: _, account: _}), do: {:error, :kyc_without_address}
+  def create_kyc(%{address: _, account: _}), do: {:error, :address_without_kyc}
+
+  # SignUp without KYC
+  def create_kyc(params), do: params
```
lucca65, line 183: this will make sure we skip KYC flow if previous function don't match. Its required that the user fill both `kyc` and `address`, only one of them will result in an error

lib/cambiatus/auth/sign_up.ex
```diff
@@ -0,0 +1,213 @@
...
+
+  def invite_user({:error, _} = error), do: error
+  def invite_user({:error, _, _} = error), do: error
```
lucca65, line 186: those type of matches are for the pipeline to just ignore and keep going if an error happens

lib/cambiatus/eos.ex
```diff
@@ -18,59 +18,46 @@ defmodule Cambiatus.Eos do
 
   require Logger
 
-  @spec unlock_wallet() :: :ok | :error
-  def unlock_wallet do
+  @spec unlock_wallet() :: boolean()
```
lucca65, line 21: I've changed this to be a boolean just because dialyzer didn't liked using `:ok` atoms without a tuple to hold it

lib/cambiatus/eos.ex
```diff
@@ -18,59 +18,46 @@ defmodule Cambiatus.Eos do
...
 
       {:error, _error} ->
-        :error
+        false
     end
   end
 
-  def create_account(account, key) do
-    create_account(%{"account" => account, "ownerKey" => key, "activeKey" => key})
-  end
```
lucca65, line 40: this function was here to support the old way of calling this. Removed

lib/cambiatus/kyc/kyc.ex
```diff
@@ -27,6 +23,16 @@ defmodule Cambiatus.Kyc do
...
   end
 
+  def create(account, kyc, address) do
+    Multi.new()
+    |> Multi.insert(:kyc, KycData.changeset(%KycData{}, Map.merge(kyc, %{account_id: account})))
+    |> Multi.insert(
+      :address,
+      Address.changeset(%Address{}, Map.merge(address, %{account_id: account}))
+    )
+    |> Repo.transaction()
```
lucca65, line 33: This will make sure the insert rolls back in case something is off

lib/cambiatus/kyc/kyc_data.ex
```diff
@@ -124,16 +132,16 @@ defmodule Cambiatus.Kyc.KycData do
...
-        ~r/[1-9]{1}\d{9}/
+        ~r/^[1-9]{1}\d{9}$/
 
       "mipyme" ->
-        ~r/\d-?\d{3}-?\d{6}/
+        ~r/^\d-?\d{3}-?\d{6}$/
 
       "gran_empresa" ->
-        ~r/\d-?\d{3}-?\d{6}/
+        ~r/^\d-?\d{3}-?\d{6}$/
```
lucca65, line 144: I've noticed that a valid document followed by some random string would make this match, so I've added the anchors

lib/cambiatus_web/controllers/chain_controller.ex
```diff
@@ -11,10 +10,4 @@ defmodule CambiatusWeb.ChainController do
     {:ok, response} = Chain.get_info()
     json(conn, response.body)
   end
-
-  def create_account(conn, params) do
-    with {:ok, result} <- Eos.create_account(params) do
-      json(conn, result)
-    end
-  end
```
lucca65, line 19: we don't support creating account this way anymore

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -26,23 +26,20 @@ defmodule CambiatusWeb.Schema.AccountTypes do
...
+      arg(:name, non_null(:string))
+      arg(:account, non_null(:string))
+      arg(:email, non_null(:string))
+      arg(:public_key, non_null(:string))
+      arg(:user_type, non_null(:string))
+      arg(:invitation_id, :string)
+      arg(:kyc, :kyc_data_update_input)
+      arg(:address, :address_update_input)
+
+      resolve(&Accounts.sign_up/3)
```
lucca65, line 39: I figured removing the `input` nesting types would make everything easier to understand without having to navigate the schema to find the matching type.

test/cambiatus_web/schema/resolvers/accounts_test.exs
```diff
@@ -4,46 +4,9 @@ defmodule CambiatusWeb.Schema.Resolvers.AccountsTest do
...
-      }
-      """
-
-      res = conn |> post("/api/graph", query: query, variables: variables)
-
-      response = json_response(res, 200)
-
-      assert response["data"]["signUp"]["status"] == "SUCCESS"
-      assert response["data"]["signUp"]["reason"] == ""
-    end
```
lucca65, line 45: I've removed the resolver test because it is unnecessary. To be honest the resolvers that we use on the project don't link different types of data, so we might as well remove all. I didn't did this in this PR because some modules are only tested on the domain, not on the resolver; and others, only in the resolver, not in the domain (testing stuff on Cambiatus and not CambiatusWeb)

I feel we should go to the direction of having the domain fully tested, and the resolvers only tested if necessary (ie, it goes through the domain and somewhere else to resolve the data). Right now the app feels very repetitive and bureaucratic. I feel we should simply it

lib/cambiatus/auth/sign_up.ex
lucca65: also if we ever want to add a kind of auth token, we can do so by adding a new step and changing the result

amiskov (commented):
Yay, one step further towards the better sign up process! I left some comments here and there.

@lucca65 BTW, since I created this PR I can't approve it nor request changes. I guess we will need @victorgdev 's help here :)

lucca65:
@amiskov commits sent with the KYC rule fix

lib/cambiatus/commune/action.ex
```diff
@@ -25,6 +21,8 @@ defmodule Cambiatus.Commune.Action do
     field(:has_proof_code, :boolean, default: false)
     field(:photo_proof_instructions, :string)
 
+    field(:position, :integer)
```
amiskov, line 24: @lucca65 Can you please explain the purpose of the `position` field? I just don't get it...

lucca65: @amiskov this is related to this conversation on #support-verdes https://cambiatus.slack.com/archives/G01GNQB5G5B/p1609968128028400 just follow the conversation

Basically its about the ordering on which the actions show under an objective. We already have a screen on Figma to handle this, but its not yet fully implemented, this is the basic implementation of this, just so Verdes can go on

lib/cambiatus/kyc/kyc.ex
```diff
@@ -27,6 +23,25 @@ defmodule Cambiatus.Kyc do
     end
   end
 
+  @doc """
+  Creates new KYC and address for a given account.
```
amiskov, line 27: This particular function clause doesn't use the address, only KYC.

```suggestion
  Creates a new KYC for a given account.
```

amiskov (commented):
Looks good! I found only one small misleading thing in a docstring.

@victorgdev can you please approve this pull-request? I can't do that because I created it some time ago.

lucca65 (approved):
@victorgdev you didn't reply so I'll ignore your review

PR #119, opened 2020-12-07
Refactor Shop Entities

lucca65:
None, but its part of a major effort to refactor our backend entities, for #113. 

Update our GraphQL Shop schemas do follow changes introduced by the past few months

- Rename tables 
- Add a new context for Shop
- Add a new type module only for shop operations
- Remove old/unused shop queries
- Start using reusable queries
- Added two new community counters

- Run `mix test`.
- Use all the frontend features necessary, such as creating new Products, viewing your notification history for orders, check the listing of shop


lib/cambiatus/shop/order.ex
```diff
@@ -1,24 +1,28 @@
...
     field(:units, :integer)
 
+    field(:created_block, :integer)
+    field(:created_tx, :string)
+    field(:created_eos_account, :string)
+    field(:created_at, :utc_datetime)
+
     belongs_to(:community, Community, references: :symbol, type: :string)
-    belongs_to(:sale, Sale)
+    belongs_to(:product, Product)
```
amiskov, line 25: In the `orders` table the field for the product ID called `sale_id`. So, we have to rename the column to `product_id` to use `belongs_to` this way or use `foreign_key: :sale_id` parameter. I think renaming is the way to go.

amiskov (changes requested):
The code looks much cleaner now! I'm very happy that you are doing this refactoring :)

Found a couple of things to fix in the `Order` module. I found them by visiting the Notification page (the bell at the top right) from the related [frontend branch](https://github.com/cambiatus/backend/pull/119). The backend couldn't find `product_id`.

lib/cambiatus/shop/order.ex
amiskov: It will raise an error while trying to request orders e.g. from the Notification page.

lucca65: Nice catch @amiskov! thanks for noticing that! I've renamed the column

lucca65:
@amiskov I've added a few more queries for some fixes I'm doing based on an user report, I thought that the work of having several PRs for those things would be too much trouble.

amiskov:
@lucca65 It looks like we also need to update the event-source since it uses the `db.sales` table for inserting/updating products. E.g. in [this file](https://github.com/cambiatus/event-source/blob/master/src/updaters/community.js).

You can check the bug by creating/updating the product in the Shop. There will be nothing changed on the frontend and if you check the `pm2 logs event-source` on Staging you'll see an error pointing to the [updaters/community.js](https://github.com/cambiatus/event-source/blob/master/src/updaters/community.js#L193) from the `event-source` repo.

lucca65:
sure, it will be done in the event-source repo!

lucca65:
done and fixed on staging: https://github.com/cambiatus/event-source/pull/29

amiskov (approved):
Tudo bem para mim! 👍

PR #122, opened 2021-01-15
Feature/search

### LINKED ISSUE #120: Implement Search
It will be a new GraphQL Object that for now only searches for the shop, it should have a structure to differentiate from other results later

It will not be paginated and will always take two params: **communityId** and **query**

lucca65:
Closes #120

- Adds a new type, search, that allows to quickly fuzzy search several different entities on Cambiatus
- Search supports actions and products
- It adds performance indexes, so its always fast
- It also uses pg extensions to reach fuzzy search

N/A


lib/cambiatus_web/schema/search_types.ex
```diff
@@ -0,0 +1,31 @@
...
+    field(:products, non_null(list_of(non_null(:product)))) do
+      arg(:query, :string)
+      resolve(dataloader(Cambiatus.Shop))
+    end
+
+    field(:actions, non_null(list_of(non_null(:action)))) do
+      arg(:query, :string)
+      resolve(dataloader(Cambiatus.Commune))
+    end
+  end
```
amiskov, line 30: With this API we will have to send the search query from the frontend twice, like this:

```graphql
query {
  search(communityId: "0,AMTA") {
    actions(query: "magic mushrooms") {
      description
    }
    products(query: "magic mushrooms") {
      title
    }
  }
}
```

I think it'll work just fine but wouldn't it be better to have it like this?

```graphql
query {
  search(communityId: "0,AMTA", query: "test") {
    actions {
      description
    }
    products {
      title
    }
  }
}
```

lucca65: yeah m! there are two reasons to it to be like this:

1) to be composable and perform better, if not we would not be able to use data loader and would have to fetch all the nested fields even if they weren't added to the query, or at least i couldn't find a way to avoid that
2) to be more flexible and usable from other screens. since it uses expensive resources of the database to do this query, i feel we shouldn't keep adding Product |> Product.search in more places, so this query is flexible to be used while not only in the global search

amiskov (approved):
Nice!

PR #123, opened 2021-01-18
Fix/sign up schema

lucca65:
Closes N/A

- Adjust schema to allow for having a profile back from the sign up, we were having match errors because of that

- mix test
- and try to sign up


lib/cambiatus_web/schema/account_types.ex
```diff
@@ -64,7 +64,8 @@ defmodule CambiatusWeb.Schema.AccountTypes do
 
   object :sign_up_response do
     field(:status, non_null(:sign_up_status))
-    field(:reason, non_null(:string))
+    field(:reason, :string)
+    field(:profile, :profile)
```
amiskov, line 68: Looks like GraphQl types should be also updated on the frontend.

lib/cambiatus_web/resolvers/accounts.ex
```diff
@@ -41,8 +41,8 @@ defmodule CambiatusWeb.Resolvers.Accounts do
         Sentry.capture_message("Sign up failed", extra: %{error: reason})
         {:ok, %{status: :error, reason: reason}}
 
-      {:ok, result} ->
-        {:ok, %{status: :success, reason: result}}
+      {:ok, user} ->
+        {:ok, %{status: :success, reason: "ok", profile: user}}
```
amiskov, line 45: Maybe we don't need `reason` for the success branch? It seems reasonable to have the reason only for the `:error`.

amiskov (changes requested):
Found a couple of places to improve. Please, check the comments!

Also, we have broken test in `test/cambiatus/auth/sign_up_test.exs:258`. Seems like this clause is unnecessary and even wrong (it breaks the test):
https://github.com/cambiatus/backend/blob/10a216bd48baeda7bec8c05a4e464ac3cc7ec9a0/lib/cambiatus/auth/sign_up.ex#L202

We actually should save KYC without an Address for `natural` users and this case is covered above the line `202`:
https://github.com/cambiatus/backend/blob/10a216bd48baeda7bec8c05a4e464ac3cc7ec9a0/lib/cambiatus/auth/sign_up.ex#L185-L190

lib/cambiatus_web/schema/account_types.ex
lucca65: yeah! we always require an update when we changing the schema

lib/cambiatus_web/resolvers/accounts.ex
lucca65: its already optional on the schema, but it will error out if we don't fill it and the user requests it

amiskov (approved):
Looks good!

PR #125, opened 2021-02-02
Feature/profile contacts

### LINKED ISSUE #121: Add the ability to add your phone number
Use the existing update profile mutation.

Also find a way to check if the user has Telegram and/or WhatsApp and flag it somehow

lucca65:
Closes #121 

- Add contacts API to add data to profile


lib/cambiatus/error.ex
```diff
@@ -0,0 +1,35 @@
...
+    |> Ecto.Changeset.traverse_errors(fn {err, opts} ->
+      Enum.reduce(opts, err, fn {key, value}, acc ->
+        String.replace(acc, "%{#{key}}", to_string(value))
+      end)
+    end)
+  end
+
+  def from(other) do
+    Logger.error("Unhandled error term: \n#{inspect(other)}")
+    Sentry.capture_message("Sign up failed", extra: %{error: other})
```
amiskov, line 31: Looks like this line should be more general than `"Sign up failed"`. Maybe like this?

```suggestion
    Sentry.capture_message("Unhandled error term:", extra: %{error: other})
```

lib/cambiatus/error.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Cambiatus.Error do
+  @moduledoc """
+  Standard error structure to be used in Cambiatus.
+
+  Used to represent and display operation statuses in a more standard way
+
+  Errors have two fields:
+  * `type`: atom representing the possible error types
+  * `message`: strings with the error message
+  """
```
amiskov, line 10: Wow! Looks very promising 👍

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -3,73 +3,92 @@ defmodule CambiatusWeb.Schema.AccountTypes do
...
+      )
+
+      arg(:invitation_id, :string,
+        description: "Optinal, used to auto invite an user to a community"
+      )
+
+      arg(:kyc, :kyc_data_update_input, description: "KYC data")
+      arg(:address, :address_update_input, description: "Address data")
+
+      resolve(&AccountsResolver.sign_up/3)
```
amiskov, line 57: Really nice to have those description fields!

amiskov (changes requested):
I really like these changes!
While checking updates on the GraphQl schema I tried to generate types on the frontend with `elm-graphql` CLI and it froze. Then I tried to check `http://localhost:4000/api/graphiql` and it sad that there's no schema available. @lucca65 can you please check this wired thing on your side?

lib/cambiatus/error.ex
lucca65: ops, thanks for noticing bro 😀

lucca65:
Found it! turns out we [cannot use objects as input](https://github.com/graphql/graphql-js/issues/151)! Thanks for testing this out before merge @amiskov

amiskov (approved):
Looks "todo bem"! 👍

PR #126, opened 2021-02-07
Feature/316 add auth

### LINKED ISSUE #61: Add Authentication to the API

lucca65:
Closes #61 

- [x] Added SignIn to GraphQL
- [x] Added Authentication for usage of our GraphQL API
- [x] Scoped some data to be only available for the current logged User
- [x] Improved documentation
- [x] Improve search query to include only available actions and available products
- [x] Add new orders query (inside product and community)

**All queries and mutations are affected**, now all users are required to be logged to query our API. This also means that some queries that required an identification no longer requires it, so it causes little changes here and there if necessary.

 here is a full list of them:

### Queries
 - [x] Claim
 - [x] Claim Analysis
 - [x] Claim Analysis History
 - [x] Communities
 - [x] Community
 - [x] Country
 - [x] Notification History
 - [x] Objective
 - [x] Product
 - [x] Products
 - [x] Search
 - [x] Transfer
 - [x] User

### Mutations
- [x] Sign In
- [x] Sign Up
- [x] Update User
- [x] Upsert KYC
- [x] Upsert Address
- [x] Delete KYC
- [x] Delete Address
- [x] Register Push
- [x] Read Notification
- [x] Complete Objective

- run `mix test`
- Do a full test using the app, focusing on Sign In, Sign Up and invitations flows


lucca65:
@venomnert can you also take a look at this?

lib/cambiatus/commune/action.ex
```diff
@@ -65,6 +65,18 @@ defmodule Cambiatus.Commune.Action do
     |> where([a], a.is_completed == ^is_completed?)
   end
 
+  def available(query \\ Action) do
+    {:ok, now} = DateTime.now("Etc/UTC")
+
+    query
+    |> join(:inner, [a], obj in Objective, on: obj.id == a.objective_id)
+    |> completed(false)
+    |> where([a], a.verification_type == "claimable")
```
amiskov, line 74: @lucca65 Just found out that we have a broken test in `test/cambiatus/commune/commune_test.exs:170`. The test uses the `action_factory` for creating sample actions with random `verification_type` (claimable or automatic).

lucca65: thanks for noticing this @amiskov! Fixed!

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -51,23 +58,38 @@ defmodule CambiatusWeb.Schema.AccountTypes do
...
+    )
+
+    field(:bio, :string, description: "Optional, short bio to let others know more about you")
+    field(:location, :string, description: "Optional, location, can be virtual or real")
+    field(:interests, :string, description: "Optional, a list of strings interpolated with `-`")
+    field(:avatar, :string, description: "Optional, URL that must be used as an avatar")
+
+    field(:contacts, list_of(non_null(:contact_input)),
+      description:
+        "Optional, list sent here will substitute all other entries, always send everything"
```
venomnert, line 91: @lucca65 can we update the wording to reflect the accurate behaviour: "Optional, list will overwrite all entries, ensure to send everything"

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -51,23 +58,38 @@ defmodule CambiatusWeb.Schema.AccountTypes do
         description: "Optinal, used to auto invite an user to a community"
```
venomnert, line 58: @lucca65 spelling error: "Optional"

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -42,7 +47,9 @@ defmodule CambiatusWeb.Schema.AccountTypes do
         description: "EOS Account public key, used for creating a new account"
       )
 
-      arg(:user_type, non_null(:string),
+      arg(:password, non_null(:string))
+
+      arg(:user_type, :string,
```
venomnert, line 52: @lucca65 I believe :user_type is a required field. 
```
arg(:user_type, non_null(:string), description: "User type informs if its a 'natural' or 'juridical' user for regular users and companies")
```

lib/cambiatus_web/schema/account_types.ex
lucca65: nice!! thanks!

lib/cambiatus_web/schema/account_types.ex
lucca65: nice catch!

test/cambiatus_web/schema/resolvers/accounts_test.exs
```diff
@@ -7,10 +7,12 @@ defmodule CambiatusWeb.Schema.Resolvers.AccountsTest do
   alias Cambiatus.{Accounts.User, Commune.Transfer}
 
   describe "Accounts Resolver" do
-    test "collects a user account given the account name", %{conn: conn} do
+    test "collects a user account given the account name" do
       assert Repo.aggregate(User, :count, :account) == 0
```
venomnert, line 11: @lucca65 can you let me the reasoning for this assert statement?

test/cambiatus_web/schema/resolvers/accounts_test.exs
```diff
@@ -7,10 +7,12 @@ defmodule CambiatusWeb.Schema.Resolvers.AccountsTest do
...
       usr = insert(:user)
```
venomnert, line 12: @lucca65 can we be consistent with the naming of the user variable, some test cases use `usr` and others use `users`.

venomnert: Please check for all other schema tests: `commune_test.exs` and `kyc_test.exs`

lib/cambiatus/shop/order.ex
```diff
@@ -16,7 +16,6 @@ defmodule Cambiatus.Shop.Order do
     field(:created_eos_account, :string)
     field(:created_at, :utc_datetime)
 
-    belongs_to(:community, Community, references: :symbol, type: :string)
```
amiskov, line 19: @lucca65 looks like after removing this relation we need to update the `:order` object from GraphQL schema since it has to be `non_null`:

https://github.com/cambiatus/backend/blob/e3674970b03ab25e08da0fa2fe8b8765b8e0b8d8/lib/cambiatus_web/schema/shop_types.ex#L64-L67

test/cambiatus_web/schema/resolvers/accounts_test.exs
lucca65: just making sure the data mutates as expected during the calls

test/cambiatus_web/schema/resolvers/accounts_test.exs
lucca65: ok, fixed!

lib/cambiatus/shop/order.ex
lucca65: wow, really nice catch @amiskov thanks for that!

venomnert (approved):

PR #129, opened 2021-02-18
Fix/fix phone regex

### LINKED ISSUE #128: Fix phone validation regex
Fix the phone regex for phone/whatsapp contact validation in `lib/cambiatus/accounts/contact.ex` to support brazilian numbers in the format +55 xx 9xxxxxxxx

lucca65:
Closes #128 

- Adjust phone regex as suggested by @NeoVier  
- Resolved regex for social media users
- Standardized international phone number validation

- run `mix test`


lib/cambiatus/accounts/contact.ex
```diff
@@ -29,15 +29,14 @@ defmodule Cambiatus.Accounts.Contact do
     regex =
       case changeset.params["type"] do
         "phone" ->
-          ~r/^[\+]?[(]?[0-9]{3}[)]?[-\s\.]?[0-9]{3}[-\s\.]?[0-9]{4,6}$/
+          ~r/^\+?\(?[0-9]{3}\)?[-\s\.]?\(?[0-9]{2}?\)?[-\s\.]?[0-9]{3,5}[-\s\.]?[0-9]{3,4}$/
 
         "whatsapp" ->
-          ~r/^[\+]?[(]?[0-9]{3}[)]?[-\s\.]?[0-9]{3}[-\s\.]?[0-9]{4,6}$/
+          ~r/^\+?\(?[0-9]{3}\)?[-\s\.]?\(?[0-9]{2}?\)?[-\s\.]?[0-9]{3,5}[-\s\.]?[0-9]{3,4}$/
```
amiskov, line 35: Regexes are hard! What about being super explicit here? Something like

```elixir
    # Mathes
    # Brazil: +(055) (47) 98849-4288
    # Costa-Rica: +506 12345678
    # Ethiopia: +251 12 345 6789
    br_cr_eth_phone =
      ~r/^\+?\(?[0-9]{3}\)?[-\s\.]?\(?[0-9]{2}?\)?[-\s\.]?[0-9]{3,5}[-\s\.]?[0-9]{3,4}$/

    regex =
      case changeset.params["type"] do
        "phone" ->
          br_cr_eth_phone

        "whatsapp" ->
          br_cr_eth_phone
```

Also, probably, it's better to validate phone numbers separately, depending on the country. I don't think that this is the case for now, but we may consider this when we'll have more countries involved.

NeoVier: Validating based on country might be a better idea for the future, especially if we care about parentheses, dashes, spaces, etc. Right now, we only [support USA](https://cambiatus.slack.com/archives/G01GNQB5G5B/p1614097747007200) numbers without punctuation, so a number like "+001 1234567890" works, but people might actually write them like ["+001 (123) 456-7890"](https://howtophoneto.com/united-states/) (but the frontend could sanitize the input and remove punctuation, just to be sure). Juggling all of these countries, while supporting punctuation will get harder and harder with time.

venomnert: @amiskov and @NeoVier regarding phone number validation for different countries (or US) we can leverage an existing elixir library: https://github.com/socialpaymentsbv/ex_phone_number

cc: @lucca65

lucca65: Thanks for the inputs folks, I really love the idea of us validating numbers by country and using a standard way of doing so. Lets advance on this front, using the google based libraries

lib/cambiatus/accounts/user.ex
```diff
@@ -54,7 +54,8 @@ defmodule Cambiatus.Accounts.User do
     |> unique_constraint(:account)
     |> validate_format(:email, ~r/@/)
     |> validate_format(:account, ~r/^[a-z1-5]{12}$/)
-    |> assoc_contacts(attrs)
+    |> cast_assoc(:contacts, with: &Contact.changeset/2)
+    # |> assoc_contacts(attrs)
```
amiskov, line 58: Looks like this commented line is redundant.

amiskov (approved):
Nice! I think now the code looks more straightforward.

I added one minor comment and also there's a minor [warning](https://github.com/cambiatus/backend/pull/129/files#diff-5a45e9e847b66cdedf9425080ce7b4869c72a63a9b0d28904246dba79f8ebeaaR12) from Credo.

lib/cambiatus/accounts/user.ex
venomnert: We can also remove `assoc_contacts/2` function as well.

lib/cambiatus/accounts/user.ex
```diff
@@ -54,7 +60,7 @@ defmodule Cambiatus.Accounts.User do
     |> unique_constraint(:account)
     |> validate_format(:email, ~r/@/)
     |> validate_format(:account, ~r/^[a-z1-5]{12}$/)
-    |> assoc_contacts(attrs)
```
venomnert, line 57: We can remove `assoc_contacts/2` function.

venomnert (approved):

lib/cambiatus/accounts/user.ex
lucca65: done!

PR #131, opened 2021-02-28
Fix/fix phone regex

### LINKED ISSUE #128: Fix phone validation regex

venomnert:

Closes #128 

- [x] Introduced a standard way to verify international phone numbers
- [x] Updated telegram and instagram username regex
- [x] Included test coverage for contacts

- Run `mix test`. Ensure all test pass
- Create a contact info of `phone` type and update it `whatsapp`.
- Try entering invalid telegram and instagram ids.

Ensure to put focus on user's contact creation.


lib/cambiatus/accounts/contact.ex
```diff
@@ -26,24 +33,31 @@ defmodule Cambiatus.Accounts.Contact do
...
+        validate_format(changeset, :external_id, ~r/(?s).*/)
+    end
+  end
 
-        _ ->
-          ~r/(?s).*/
-      end
+  defp validate_phone_number(changeset, country_code \\ @default_country_code) do
+    external_id = changeset.changes.external_id
+    {:ok, phone_number} = ExPhoneNumber.parse(external_id, country_code)
```
lucca65, line 56: this way we are always checking the number as a US number, maybe we can try to pattern match / regex match the prefix to get the number? maybe expect always a `(\+){1}\d{2}` for the prefix

lib/cambiatus/accounts/contact.ex
```diff
@@ -26,24 +33,31 @@ defmodule Cambiatus.Accounts.Contact do
...
 
-    validate_format(changeset, :external_id, regex)
+    case ExPhoneNumber.is_possible_number?(phone_number) do
```
lucca65, line 58: I really think cases of true/false should be `if`s

lucca65 (changes requested):

lib/cambiatus/accounts/contact.ex
NeoVier: shouldn't the prefix be `\+?\d{1,3}`? There are country codes that are 1 digit long (US), 2 digits long (BR), and 3 digits long (CR)

lucca65: +1 but the `+` shouldn't be optional

venomnert: I believe the library can automatically detect phone number and its country. In that case we probably won't need regex.

I need to dig more into this.

lib/cambiatus/accounts/contact.ex
venomnert: I have made the update @lucca65

lib/cambiatus/accounts/contact.ex
venomnert: Even though I set the default country code `BR` the library parses and recognizes all valid countries. The current implementation recognizes CR, US and BR and other countries, so no change is required.

lib/cambiatus/accounts/contact.ex
```diff
@@ -26,24 +33,32 @@ defmodule Cambiatus.Accounts.Contact do
   end
 
   def validate_external_id(changeset) do
-    regex =
-      case changeset.params["type"] do
-        "phone" ->
-          ~r/^\+?\(?[0-9]{3}\)?[-\s\.]?\(?[0-9]{2}?\)?[-\s\.]?[0-9]{3,5}[-\s\.]?[0-9]{3,4}$/
+    case changeset.changes.type do
```
lucca65, line 36: uh I didn't knew about this! nice!

lib/cambiatus/accounts/contact.ex
lucca65: oh I see! nice!

lucca65 (approved):

venomnert:
- Resolved regex for social media users
- Standardized international phone number validation

PR #133, opened 2021-03-23
Feature/35 support subdomain

### LINKED ISSUE #132: Implement multi domain support
We need to support multi domains on our backend. 

- Communities should be able to choose a subdomain (we need to have a blacklist filled from the migration)
- Backend should take the host into consideration when receiving requests

#### Context
https://github.com/cambiatus/improvement-proposals/issues/35

venomnert:

Closes #132

* Allow a community to have their own `subdomain`
  * This is an optional field and can only be added by admin
* Subdomains can't be any of the following reserved word `dev, demo, staging`

* Run `mix test` you should get a success message.

### DIFF
--- a/lib/cambiatus/commune/community.ex
+++ b/lib/cambiatus/commune/community.ex
@@ -52,7 +52,7 @@ defmodule Cambiatus.Commune.Community do
 
   @required_fields ~w(symbol creator name description inviter_reward invited_reward)a
   @optional_fields ~w(logo type supply max_supply min_balance issuer precision created_block
-  created_tx created_at created_eos_account)a
+  created_tx created_at created_eos_account subdomain)a
 
   @doc false
   def changeset(%Community{} = community, attrs) do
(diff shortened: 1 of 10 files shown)


lib/cambiatus/commune/community.ex
```diff
@@ -13,12 +13,14 @@ defmodule Cambiatus.Commune.Community do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @reserved_subdomain ~w(dev demo staging)
```
lucca65, line 16: I know this is still a draft, but what do you think about us putting this into a migration? So its more flexible for us to add more of those used words? I mean, people will for sure use weird words here, so maybe its best to have a more flexible way of adding names without having to redeploy stuff

lib/cambiatus_web/subdomain_router.ex
```diff
@@ -0,0 +1,18 @@
...
+  pipeline :api do
+    plug(:accepts, ["json"])
+  end
+
+  scope "/", CambiatusWeb do
+    pipe_through(:api)
+
+    get("/", Subdomainer.Subdomain.PageController, :index)
+  end
+end
```
lucca65, line 18: why we need a new router?

venomnert:
@lucca65 closing this because of PR #149

PR #136, opened 2021-03-30
Feature/add claim analysis direction

lucca65:
## What issue does this PR close
N/A

## Changes Proposed ( a list of new changes introduced by this PR)
- Add the ability to define the sort order for the Claim Analysis History query

## How to test ( a list of instructions on how to test this PR)
- mix test

Issue #127, opened 2021-02-11
Improve Authentication


lucca65:
Add the following features:

- Sessions table, storing expiration, last time requested and validity. Use that to serve and validate session tokens. 
- Add validation of message. Client signs a suggested message with their private key and we validate its ownership using the account's public key. Maybe NIFs are the way to go
- Store failed login attempts
- Add auto-invite configurations migrations / mutations
- Adjust the signUp function to respect that auto invite, always using a) the invitation b) the request header's url

For session tokens we must have a way to invalidate certain user access to the database

- [ ] Add testing
- [ ] Security audit
- [x] Update elixir #49

PR #140, opened 2021-04-06
Session table

venomnert:
* This PR is a milestone to major branch #137 

* Utilize session table to stores information during sign in process: phrase, token, account, app version, etc... 

1. `mix test`
2. Run the following queries: `startAuthSession`, `signInV2`, `signOut`


lib/cambiatus/auth/auth.ex
```diff
@@ -64,6 +71,126 @@ defmodule Cambiatus.Auth do
     end
   end
 
+  def gen_auth_phrase(user) do
+    user
+    |> create_auth()
+    |> create_session()
+  end
+
+  def create_auth(user) do
```
lucca65, line 80: maybe `create_phrase` because its what the function return?

lib/cambiatus/auth/auth.ex
```diff
@@ -64,6 +71,126 @@ defmodule Cambiatus.Auth do
...
+    end
+  end
+
+  def verify_signature(account, signature) do
+    result =
+      case get_user_token(account, :auth) do
+        %{phrase: phrase} ->
+          account
+          |> Ecdsa.verify_signature(signature, phrase)
+          |> case do
```
lucca65, line 128: in those cases we can use `if`, pattern matching booleans 🤔

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -80,6 +80,12 @@ defmodule CambiatusWeb.Schema.AccountTypes do
       resolve(&AccountsResolver.sign_in/3)
     end
 
+    field :sign_out, :string do
+      arg(:account, non_null(:string))
+
+      resolve(&AccountsResolver.sign_out/3)
```
lucca65, line 86: This should be an `arg`less mutation, and it should be authenticated. The way its described, looks like anyone can signOut a given `account`, if its a authenticated API call, then we already will have the user on the context

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -23,12 +23,12 @@ defmodule CambiatusWeb.Schema.AccountTypes do
     end
 
     @desc "Sign in phrase"
-    field :get_auth_phrase, :auth_phrase do
+    field :start_auth_session, :auth_session do
```
lucca65, line 26: just a suggestion, I'm not 100% confident about this, but maybe `generate_auth` or `gen_auth`, or `init_session` `generate_session`, etc

lucca65 (changes requested):

lib/cambiatus/auth/auth.ex
venomnert: I have made the udpate

lib/cambiatus/auth/auth.ex
venomnert: Agreed and made the update. I will need to ensure to follow this pattern for future changes.

lib/cambiatus_web/schema/account_types.ex
venomnert: That make sense. I have updated it so only the authorized user can logout.

lib/cambiatus_web/schema/account_types.ex
venomnert: Updated to `gen_auth`

lib/cambiatus_web/resolvers/commune.ex
```diff
@@ -63,6 +63,20 @@ defmodule CambiatusWeb.Resolvers.Commune do
...
+        args
+        |> Map.get(:filter)
+        |> Enum.reduce(query, fn
+          {:direction, direction}, query ->
+            Claim.ordered(query, direction)
+        end)
+      else
+        query
+        |> Claim.ordered(:desc)
+      end
```
lucca65, line 78: is your branch updated with master?

venomnert: It is. Am i missing something?

lucca65: but did you added this? the claim ordering was released here: https://github.com/cambiatus/backend/pull/139/files and here:https://github.com/cambiatus/backend/pull/136/files

venomnert: I believe the update has been made now.

lib/cambiatus/auth/auth.ex
```diff
@@ -44,26 +51,154 @@ defmodule Cambiatus.Auth do
...
+      end
+
+    delete_user_token(%{account: account, filter: :auth})
+    result
+  end
+
+  def get_user_token(%{account: account, filter: :session}) do
+    from("user_tokens",
+      where: [context: "session", user_id: ^account],
+      select: [:context, :user_id, :token]
```
lucca65, line 148: those two lines get repeated a lot, why not move it to UserToken module so the code get less repetitive, functions like `by_account` or `by_token`, etc.

lib/cambiatus/auth/ecdsa.ex
```diff
@@ -4,74 +4,35 @@ defmodule Cambiatus.Auth.Ecdsa do
...
+    pub_key =
+      account
+      |> EosjsAuthWrapper.get_account_info()
+      |> get_pub_key()
 
-        signature_elixir = %EllipticCurve.Signature.Data{
-          r: String.to_integer(signature_points["r"]),
-          s: String.to_integer(signature_points["s"])
-        }
+    EosjsAuthWrapper.verify(signature, phrase, pub_key)
```
lucca65, line 12: didn't this triggered credo? 

if you are on a pipe you need to start with a raw value:

signature
|> Eosjs.verify(phrase)
|> case do

lib/cambiatus/auth/ecdsa.ex
```diff
@@ -4,74 +4,35 @@ defmodule Cambiatus.Auth.Ecdsa do
...
+      {:ok, result} -> result
+      {:error, _} -> false
+    end
+  end
 
-        EllipticCurve.Ecdsa.verify?(phrase, signature_elixir, public_key) &&
-          compare_public_keys(pub_points["string"], account)
+  def sign(signature, priv_key) do
+    EosjsAuthWrapper.sign(signature, priv_key)
+  end
```
lucca65, line 21: is this really needed?

lib/cambiatus/auth/ecdsa.ex
```diff
@@ -4,74 +4,35 @@ defmodule Cambiatus.Auth.Ecdsa do
...
 
-      {:error, _} ->
-        false
-    end
+  def sign_with_random(phrase) do
+    EosjsAuthWrapper.gen_rand_signature(phrase)
```
lucca65, line 24: same here, if the objetive is to have less lengthy words you can use aliases

lucca65: alias are useful:

```elixir
alias Foo.Bar, as: Bar
```

and in the case the problem is that the function is not descriptive enough, you can rename it on the package

lib/cambiatus/auth/ecdsa.ex
```diff
@@ -4,74 +4,35 @@ defmodule Cambiatus.Auth.Ecdsa do
...
   end
 
-  @doc """
-  Verify public key associated to the signature private key and the public key for the account
-  """
-  defp compare_public_keys(public_key_a, account) do
-    {"app", :accountToPublicKey}
-    |> NodeJS.call([account])
+  defp get_pub_key(account) do
```
lucca65, line 27: not sure about this function returning nil bro, do you think we can improve it somehow?

lib/cambiatus/auth/ecdsa.ex
lucca65: also can you send a little explanation on the motivation behind random?

lib/cambiatus/auth/user_token.ex
```diff
@@ -0,0 +1,28 @@
...
+    field(:phrase, :string)
+    field(:token, :binary)
+    field(:context, :string)
+    belongs_to(:user, User, references: :account, type: :string)
+
+    timestamps(updated_at: false)
+  end
+
+  @required_fields ~w(token context)a
+  @optional_fields ~w(phrase)a
```
lucca65, line 19: maybe a blank line between those would be nice!

lib/cambiatus/commune/commune.ex
```diff
@@ -196,15 +196,18 @@ defmodule Cambiatus.Commune do
     |> join(:left, [c, a], v in Validator, on: v.action_id == c.action_id)
     |> where(
       [c, a, o, v],
-      o.community_id == ^community_id and v.validator_id == ^account and a.is_completed == false and
+      o.community_id == ^community_id and
+        v.validator_id == ^account and
+        a.is_completed == false and
+        not (a.usages > 0 and a.usages_left == 0) and
+        (is_nil(a.deadline) or fragment("NOW()") < a.deadline) and
```
lucca65, line 203: I'm confused about those lines, did you added them?

lib/cambiatus_web/resolvers/accounts.ex
```diff
@@ -49,15 +59,13 @@ defmodule CambiatusWeb.Resolvers.Accounts do
...
-        CambiatusWeb.AuthToken.invalidate(token)
-        {:error, message: "Sign In failed", details: Cambiatus.Error.from(reason)}
-
-      {:ok, user} ->
-        {:ok, %{user: user, token: CambiatusWeb.AuthToken.sign(user)}}
+    with {:ok, _token_data} <- Auth.verify_session_token(account, token),
+         {:ok, user} <- Auth.verify_signature(account, signature) do
+      {:ok, %{user: user, token: token}}
+    else
+      error -> error
```
lucca65, line 68: this will match anything, maybe you want to 

```
else
  {:error, details} = error -> error
end

File: lib/cambiatus_web/resolvers/commune.ex (hunk 5 above)
Comment (lucca65) [reply lib/cambiatus_web/resolvers/commune.ex:78]: yeah did you wrote those lines? they look from the PR we did a while a go

File: mix.exs [hunk 14]
```diff
@@ -48,7 +48,6 @@ defmodule Cambiatus.Mixfile do
       {:uuid, "~> 1.1"},
       {:magic_number, "~> 0.0.4"},
       {:mogrify, "~> 0.8.0"},
-      {:nodejs, "~> 2.0"},
```
lucca65, line 51: phew 🌶️

lucca65 (changes requested):
Good! we are on the right direction! I'm missing the tests attesting the new signUp and signIn flow from end to end

lib/cambiatus/auth/ecdsa.ex
lucca65: credo should be ruined automatically by GitHub, but you can always run `mix credo`! 💪

lib/cambiatus/auth/ecdsa.ex
lucca65: also, private functions usually don't bring much benefit if you are not working on a library

lucca65: the main reason why I don't like to use private functions is that you loose stack trace if you run into an error on them

lib/cambiatus_web/resolvers/accounts.ex
venomnert: Made the update

lib/cambiatus/auth/user_token.ex
venomnert: Made the necessary format

lib/cambiatus/auth/auth.ex
venomnert: I have extracted this into it's own module `Auth.Session`

lib/cambiatus_web/resolvers/commune.ex
venomnert: As discussed on our call this branch will update the main feature branch `improve-auth`.

lib/cambiatus/auth/ecdsa.ex
venomnert: I have made the update

lib/cambiatus/auth/ecdsa.ex
venomnert: This used to for testing the auth process

lib/cambiatus/auth/ecdsa.ex
venomnert: > same here, if the objetive is to have less lengthy words you can use aliases
Fixed this by utilizing alias

> also can you send a little explanation on the motivation behind random?
This function is utilized to test invalid signatures

lib/cambiatus/auth/ecdsa.ex
venomnert: > not sure about this function returning nil bro, do you think we can improve it somehow?

I have made the fix for this

> the main reason why I don't like to use private functions is that you loose stack trace if you run into an error on them

I agree but the question I have is, how do we utilize helper functions in modules without polluting the api?

lib/cambiatus/commune/commune.ex
venomnert: As discussed on our call this branch will update the main feature branch improve-auth.

lib/cambiatus/auth/ecdsa.ex
lucca65: why not simply call `EosjsAuthWrapper.sign(signature, priv_key)`directly in the test too?

lib/cambiatus/auth/ecdsa.ex
lucca65: Still think that we can call `EosjsAuthWrapper` directly here, functions that alias other functions is just weird, adds a layer on something without any benefits

lib/cambiatus/auth/ecdsa.ex
lucca65: but how `defp` helps in avoiding said pollution?

lib/cambiatus/auth/ecdsa.ex
venomnert: I have made the update

lib/cambiatus/auth/ecdsa.ex
venomnert: Agreed, I made the necessary update.

lib/cambiatus/auth/ecdsa.ex
venomnert: In my opinion private functions help hide the implementation details for the main functions of the module. In this case `get_pub_key` is a helper functions and it shouldn't be called outside off this module.

.tool-versions
```diff
@@ -0,0 +1,2 @@
+elixir 1.10.3-otp-23
```
lucca65, line 1: this is weird, master is on 1.8, we have a PR to update it to 1.11

now yours is a different version too

lucca65 (approved):

lucca65:
We can do a more complete review when we are finally targeting `master`

.tool-versions
venomnert: Hey @lucca65 as discussed here https://github.com/cambiatus/backend/issues/127. This will get merged in after we update Elixir.

venomnert:
Hey @lucca65 I just added 1 more update.

I added extra tests to covers signup -> signout process

lucca65 (approved):
Looks good! ✨

PR #178, opened 2021-07-04
Feature/api improvements

### LINKED ISSUE #176: New transfer query
Create a new transfer query that accepts:

- communityId
- account
- direction: `from` or `to`
- filters: `date`, `other_account`, `min_amount`

This new query will be used on the dashboard and on the community balance tool, so it search capabilities must be available

### LINKED ISSUE #177: Add shopPreview query
We need a public query to show the following information about a sale:

- seller name
- product photo
- community name
- product name
- product description

### LINKED ISSUE #181: Adjust column size for storing text
We currently use a limit to comply with the limits of a blockchain so the text is always smaller than 256 chars long. We managed to address the limits on the contract, but we need to do the same for our Postgres tables. 

fields:

- [action] photo_proof_instructions
- [action] description
- [community] description
- [objective] description
- [product] description

lucca65:
closes #177
closes #176 
closes #181

- Improve Sentry report to SignUp so we can better understand failings when creating an account
- Add shop preview query to be used on the thumbnail
- Add user search
- Improve user transfer API to be more composable and with more filters
- Migration to adjust some database types

- mix do check, test


lib/cambiatus_web/schema/shop_types.ex
```diff
@@ -29,6 +29,13 @@ defmodule CambiatusWeb.Schema.ShopTypes do
       middleware(Middleware.Authenticate)
       resolve(&Shop.get_product/3)
     end
+
+    @desc "Public product query, mainly used to preview a sale"
+    field(:product_preview, :product_preview) do
+      arg(:id, non_null(:integer))
+
+      resolve(&Shop.get_product/3)
+    end
```
NeoVier, line 38: Can we always return a `:product_preview` here? If the product doesn't exist, the frontend receives a GraphQL error anyway, so I guess it doesn't make sense to have a `Maybe ProductPreview` on the frontend. (The same applies to the `product` query)

```suggestion
    @desc "Public product query, mainly used to preview a sale"
    field(:product_preview, non_null(:product_preview)) do
      arg(:id, non_null(:integer))

      resolve(&Shop.get_product/3)
    end
```

lucca65: opa! my bad bro

this default to null is extra clumsy intuitive to me

lib/cambiatus_web/schema/account_types.ex
```diff
@@ -101,10 +101,18 @@ defmodule CambiatusWeb.Schema.AccountTypes do
     field(:external_id, :string)
   end
 
+  input_object(:transfer_direction) do
+    field(:other_account, :string, description: "Optional other account on the transfer")
+
+    field(:direction, non_null(:transfer_direction_value),
+      description: "If the user is receiving or sending the transaction"
+    )
```
NeoVier, line 109: Could `:direction` be nullable as well? That would mean we could query for all transactions with another user, independent on direction

lucca65:
@NeoVier The duplicate problem was fixed!

lib/cambiatus_web/schema/account_types.ex
lucca65: sure, no problem, I've changed it!

NeoVier: @lucca65 if I try to query with `otherAccount`, but without `direction`, I get this error:
```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

Here's a query that doesn't work:
```
{
  user(account: "henriquebuss") {
    transfers(first:10, filter: {communityId: "0,BUSS", direction: {otherAccount: "henriquebus2"}}) {
      edges {
        node {
          amount
          communityId
          from {
            account
          }
          to {
            account
          }
        }
      }
    }
  }
}
```

lucca65: Ok! Fixed!

lucca65: and deployed on staging

PR #188, opened 2021-10-06
🗞📢 Community News

### LINKED ISSUE #179: Implement Broadcast API
# Story

I, as a Community admin
want to send news to all community members
To improve our Community access to information

### Screens

https://www.figma.com/file/De9z2EjpGRkfsEXVzfj9u2/Support-Broadcast?node-id=0%3A1

### Use cases

Given a community with at least one `news` and it has news enabled
When the users login
Then the community will have a `highlighted_news` (always only one)

Given a community with at least one `news` and it has news enabled
When a user opens it
Then the news is saved as `read` by the user

Given a community with existing `news` and it has news enabled
When someone from the community access the community page
Then we will display all the community news with
- a flag if it was read already or not
- a flag if it is a highlight or not
- a list of reactions

Given a new community
When the admin goes to settings and click on the toggle for news
Then the community has news enabled

Given a community with `news` enabled
When the admin goes to the `news` page and click on a create `news` with no schedule
Then the news become available to all users and this news is set as the highlight

Given a community with `news` enabled
When the admin goes to the `news` page and click on a `create news` with a schedule
Then the `news` will not be shown until the required date and time. After the required time has passed it will become available to all and set as the highlight

Given an existing `news` and it has news enabled
When the admin clicks on edit, change it and submit a new version
Then a new version is set

Given an open `news` and it has news enabled
When the admin clicks on any `news`
Then it shows all the versions with edition date, but only for the admin

Given an existing `news` and it has news enabled
When the admin clicks on `highlight`
Then this is set as the community highlight `news` 

### Problem description

All our communities use Cambiatus for trading and connecting, but we don't provide any tools for the admins to actually communicate with their users. Some communities do WhatsApp, some do presencial meetings. We should create some way to improve the interaction and make sure everyone gets the message. This will strengthen the network

lucca65:
Closes #179

 - Creates structs for `news`, `news_version`, and `news_reactions`
 - Add `has_news` and `highlighted_news` to the community 
 - Creates schema types for news (that also gets news_reaction)
 - Creates mutation for add news, add reactions, highlight some news, enable the community to have news, update news. 
 - Creates queries to get news(with reactions), and get a list of news within the community.
 - Creates queries to get the highlighted news
 - Add a worker to set news as highlighted according to schedule
 - Creates tests
 

`mix test`


NeoVier:
Enums are working perfectly! However, I can't get the subscription to work... If I subscribe to it and change the highlighted news, nothing happens. You can test this out on Graphiql (just change the `communityId` and `newsId`):

Subscription:
```graphql
 subscription {
  highlightedNewsChange(communityId:"0,BUSS") {
    id
    title
  }
}
```

Mutation:
```graphql
mutation {
  highlightedNews(communityId:"0,BUSS", newsId: 9) {
    highlightedNews {
      title
      id
    }
  }
}
```

flpfar:
> Enums are working perfectly! However, I can't get the subscription to work... If I subscribe to it and change the highlighted news, nothing happens. You can test this out on Graphiql (just change the `communityId` and `newsId`):
> 
> Subscription:
> 
> ```graphql
>  subscription {
>   highlightedNewsChange(communityId:"0,BUSS") {
>     id
>     title
>   }
> }
> ```
> 
> Mutation:
> 
> ```graphql
> mutation {
>   highlightedNews(communityId:"0,BUSS", newsId: 9) {
>     highlightedNews {
>       title
>       id
>     }
>   }
> }
> ```

We couldn't test it on graphiql staging (I think we need the staging password to signIn... can you help us with that?), but can you try testing the subscription with the authorization header?




> As @lucca65 mentioned during our meeting, it seems like there's an issue if we try to create news with large text. From what I've been testing, anything with length > 255 characters returns an error with status 500, and the body is just `Something went wrong`. Here is an example mutation that triggers this error (if you remove a single character from the description, it works):
> 
> ```js
> mutation {
>   news(communityId: "0,BUSS",
>     title: "256 characters",
>     scheduling: null,
>     description:"1iam quam nulla porttitor massa id neque aliquam vestibulum morbi blandit cursus risus at ultrices mi tempus imperdiet nulla malesuada pellentesque elit eget gravida cum sociis natoque penatibus et magnis dis parturient montes nascetur ridiculus mus mauris"){
>      title
>     id
>   }
> }
> ```
> 
> Also, this error just came in on Slack from one of my tests: https://cambiatus.slack.com/archives/CA83HJAAD/p1635946778000100

We'll take a look on that. Thanks!

NeoVier:
> We couldn't test it on graphiql staging (I think we need the staging password to signIn... can you help us with that?)

Sure! The `signIn` mutation takes 3 parameters:

- `account` is just your username (`feliperosa12`)
- The [staging password is hardcoded as `gAK2_Fdjyy8`](https://github.com/cambiatus/frontend/blob/8aa11b4e7d421c87fd1c5862a60c63204f9d71d4/package.json#L11).
- In order to get the `invitationId`, you can log in on a community you're a member of, and click on the "Invite your friends" button on the dashboard. This will give you a link like `http://buss.staging.cambiatus.io/invite/OwYpW0`. The last part of the url (in this case, `OwYpW0`) is the `invitationId`

Once you call that mutation with these arguments, you'll get an object that has a `token`, and you can use that on the bottom left corner of GraphiQL (where it says "Http headers"), and add an `Authorization: Bearer <token>` header. Every tab that has this header is now considered authenticated

> can you try testing the subscription with the authorization header?

Yeah, I was using it already, otherwise I don't think I could even send the mutation :sweat_smile:

lib/cambiatus/workers/scheduled_news_worker.ex
```diff
@@ -16,7 +16,7 @@ defmodule Cambiatus.Workers.ScheduledNewsWorker do
       Commune.set_highlighted_news(news.community_id, news.id)
       :ok
     else
-      {:error, "deu não"}
```
lucca65: kkkkkk

lib/cambiatus/commune/commune.ex
```diff
@@ -674,4 +683,55 @@ defmodule Cambiatus.Commune do
...
+    end
+  end
+
+  defp do_set_highlighted_news({:error, _} = error, _) do
+    error
+  end
+
+  defp do_set_highlighted_news({:ok, community}, news_id) do
+    community
+    |> update_community(%{highlighted_news_id: news_id})
```
lucca65, line 716: I don't think we need a pipe here

lib/cambiatus/commune/commune.ex
```diff
@@ -674,4 +683,55 @@ defmodule Cambiatus.Commune do
...
+    else
+      {:error, "Unauthorized"}
+    end
+  end
+
+  defp do_set_highlighted_news({:error, _} = error, _) do
+    error
+  end
+
+  defp do_set_highlighted_news({:ok, community}, news_id) do
```
lucca65, line 714:
```suggestion
  defp set_highlighted_news({:ok, community}, news_id) do
```

lib/cambiatus/commune/commune.ex
```diff
@@ -674,4 +683,55 @@ defmodule Cambiatus.Commune do
...
+  def set_has_news(current_user, community_id, has_news) do
+    case get_community(community_id) do
+      {:ok, community} ->
+        if community.creator == current_user.account do
+          community
+          |> update_community(%{has_news: has_news})
+          |> case do
+            {:ok, _} = success -> success
+            {:error, _} = error -> error
+          end
```
lucca65, line 728: this case is redundant

lib/cambiatus/commune/community.ex
```diff
@@ -50,11 +51,16 @@ defmodule Cambiatus.Commune.Community do
     field(:has_shop, :boolean, default: true)
     field(:has_kyc, :boolean, default: false)
 
+    # Social
+    field(:has_news, :boolean, default: false)
+
+    belongs_to(:highlighted_news, News)
```
lucca65, line 57: I like the comment but it should use spaces to scope better visually what's social. 

```suggestion
    # Social
    field(:has_news, :boolean, default: false)
    belongs_to(:highlighted_news, News)
    
```


Or even simpler

```suggestion
    field(:has_news, :boolean, default: false)

    belongs_to(:highlighted_news, News)
```

lib/cambiatus/social/social.ex
```diff
@@ -0,0 +1,130 @@
...
+  end
+
+  def query(queryable, _params) do
+    queryable
+  end
+
+  def get_news(news_id) do
+    News
+    |> Repo.get(news_id)
+  end
```
lucca65, line 31: hmm, this call only call another function. I know alias can be useful, but here... Idk... At least we can make it shorter without the pipe:

```suggestion
  def get_news(news_id), do: Repo.get(News, news_id)
```

lib/cambiatus/social/social.ex
```diff
@@ -0,0 +1,130 @@
...
+
+          {:ok, news}
+
+        error ->
+          error
+      end
+    end
+  end
+
+  defp do_create_news(attrs) do
```
lucca65, line 45: No need to add `do`

```suggestion
  defp create_news(attrs) do
```

flpfar:
> > We couldn't test it on graphiql staging (I think we need the staging password to signIn... can you help us with that?)
> 
> Sure! The `signIn` mutation takes 3 parameters:
> 
> * `account` is just your username (`feliperosa12`)
> * The [staging password is hardcoded as `gAK2_Fdjyy8`](https://github.com/cambiatus/frontend/blob/8aa11b4e7d421c87fd1c5862a60c63204f9d71d4/package.json#L11).
> * In order to get the `invitationId`, you can log in on a community you're a member of, and click on the "Invite your friends" button on the dashboard. This will give you a link like `http://buss.staging.cambiatus.io/invite/OwYpW0`. The last part of the url (in this case, `OwYpW0`) is the `invitationId`
> 
> Once you call that mutation with these arguments, you'll get an object that has a `token`, and you can use that on the bottom left corner of GraphiQL (where it says "Http headers"), and add an `Authorization: Bearer <token>` header. Every tab that has this header is now considered authenticated
> 
> > can you try testing the subscription with the authorization header?
> 
> Yeah, I was using it already, otherwise I don't think I could even send the mutation sweat_smile

Graphql authentication worked. Thanks, @NeoVier ! We'll take a look on the subscription to see what's going on.

lib/cambiatus_web/resolvers/commune.ex
```diff
@@ -104,7 +104,7 @@ defmodule CambiatusWeb.Resolvers.Commune do
   Fetch a objective
   """
   @spec get_objective(map(), map(), map()) :: {:ok, map()} | {:error, term}
-  def get_objective(_, %{input: params}, _) do
+  def get_objective(_, params, _) do
```
lucca65, line 107: so_good

lib/cambiatus_web/schema/commune_types.ex
```diff
@@ -119,6 +119,18 @@ defmodule CambiatusWeb.Schema.CommuneTypes do
         {:ok, transfer}
       end)
     end
+
+    @desc "Subscribe to highlighted_news change"
+    field :highlighted_news_change, :news do
```
lucca65, line 124: we don't need the word `change`

```suggestion
    field :highlighted_news, :news do
```

lib/cambiatus_web/schema/commune_types.ex
```diff
@@ -119,6 +119,18 @@ defmodule CambiatusWeb.Schema.CommuneTypes do
         {:ok, transfer}
       end)
     end
+
+    @desc "Subscribe to highlighted_news change"
```
lucca65, line 123:
```suggestion
    @desc "[Auth required] Subscribe to highlighted_news change"
```

lib/cambiatus/social/social.ex
adubas: We tried to do this way but it generate conflicts with the `create_news/1` on line 30

lib/cambiatus/commune/commune.ex
lucca65: well, this function only serves as a more verbose case, right? maybe we can just use a case instead of two functions with matches

lib/cambiatus/social/social.ex
lucca65: I think its safe to simply add those three lines of this function into the original one, I see no advantage of making them separate

flpfar: Oh, we were using the `do_` from the [elixir style guide for private functions](https://github.com/christopheradams/elixir_style_guide#private-functions-with-same-name-as-public). Should we avoid using it?

lucca65: ### TL;DR: lets use less private functions if there is only one place to call that private function, makes code harder to navigate! 

**Long version**

I think the naming pattern is fine! but I don't really see why we should separate this into something private and something public in this case, since we are not trying to hide implementation details or reusing this step internally. 

I personally don't really see why we should use private functions in most of our app, because we don't have the use case where this is very useful.   

every time we create a new `news` entity, this should go through the logic of checking if it can be highlighted. We don't have a situation where we create this entity and don't do this check. Hence, we don't need the private implementation. 

In this situation we are hiding under a private function one of the most common calls in phoenix apps: repo-insert-changeset, and its not used anywhere else, only there. It may save a few lines on the future but it only makes the code harder to read now.

**Extras**

I feel that if we want to really dig into the idea of properly using private functions, we should start using a more robust API, such as [GenServer](https://hexdocs.pm/elixir/GenServer.html) that guarantees that in a nice to use and stateful API (where those pattern matched functions really start to shine and make sense, instead of a plain `case`).

I really recommend trying out the GenServer behavior, it works much like an interface/abstract class in object oriented languages, so you can just try it out someday. Since it is stateful, you can actually make it wait until you are finished with something and reply async as well! This encapsulation method is actually one of the main selling points and it allows the creation of the Actor model, a really nice "human-cell" like feature that can be used in lots of clever ways

For our use case, I think private functions, gen servers and other encapsulation methods are overkill, and sometimes makes the code harder to read as we need to go to the internal private function to check its behavior.

lib/cambiatus/commune/commune.ex
adubas: We change this function to a case, but kept the `check_user_authorization` function private.

lib/cambiatus/workers/scheduled_news_worker.ex
```diff
@@ -0,0 +1,22 @@
...
+  def perform(%Oban.Job{args: %{"news_id" => news_id, "news_scheduling" => news_scheduling}}) do
+    news = Social.get_news(news_id)
+    {:ok, time, _} = DateTime.from_iso8601(news_scheduling)
+
+    if DateTime.compare(news.scheduling, time) == :eq do
+      Commune.set_highlighted_news(news.community_id, news.id)
+      :ok
+    else
+      {:error, "Scheduling does not match"}
+    end
```
lucca65, line 20: Maybe we can, instead of checking the schedule, every time the worker runs, we can simply remove old workers (with old schedules) in favor of the most recent one. 

This way we won't have failed jobs on the list unless they failed because of something other than a new schedule.

Some docs about this matter of replacing jobs: https://hexdocs.pm/oban/Oban.html#module-unique-jobs

lib/cambiatus/social/social.ex
flpfar: I got your point! Thanks for your explanation :nerd_face: :1st_place_medal: 

I think sometimes the private functions makes the public ones easier to read/understand and can keep the code cleaner (unused private functions get a warning), but it really gets harder to navigate. 
Personally, I like them, but In these cases you pointed it makes sense to keep them inside a case or something like.

Thanks, bro!

NeoVier:
@flpfar @adubas I'm still getting an error that I think is related to the description length. All I get is "Something went wrong", when I try to create news with this description (which has 256 characters):

```
1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456
```

If i remove a single character (making it 255 characters long), it works.

As usual, the updated frontend is [here](https://deploy-preview-654--angry-wing-8fcdd0.netlify.app/), if you want to test it there (to create news, you need to be logged in with the user that created the community, then go to settings on the top right > "Communications" > "New communication")

flpfar:
> @flpfar @adubas I'm still getting an error that I think is related to the description length. All I get is "Something went wrong", when I try to create news with this description (which has 256 characters):
> 
> ```
> 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456
> ```
> 
> If i remove a single character (making it 255 characters long), it works.
> 
> As usual, the updated frontend is [here](https://deploy-preview-654--angry-wing-8fcdd0.netlify.app/), if you want to test it there (to create news, you need to be logged in with the user that created the community, then go to settings on the top right > "Communications" > "New communication")

Can you try again? I just ran the migration in staging. :+1:

NeoVier:
Nice, it's working now, thanks!

NeoVier (approved):
Sorry for the delay on the approval folks, was just waiting until the frontend PR was approved by the design team :sweat_smile:. This turned out really nice, congrats @adubas @flpfar!

PR #220, opened 2022-04-06
Feature/shop refactor

### LINKED ISSUE #221: Bug in My Claims
**Describe the bug**
in the area of ​​analyzed claims of the Play4change community, the actions that were claimed and analyzed in the other communities appear

**To Reproduce**
1. Go to my claims
2. Click on analyzed
3. See error

image

**Expected behavior**
I believe that each community should only present the actions that were created and that will be analyzed by it

lucca65:
Closes #221

- Remove blockchain data ingestion fields such as `created_tx` and `created_block` in favor of regular `inserted_at` and `updated_at`
- Add multiple photos to a single product
- Add several validations to product
- Allow products to be created and updated via GraphQL mutation
- Fix a bug where the community_id filter was not respected in the user.claim query

- Run `mix test`

### DIFF
--- a/lib/cambiatus/shop/product.ex
+++ b/lib/cambiatus/shop/product.ex
@@ -8,7 +8,8 @@ defmodule Cambiatus.Shop.Product do
   import Ecto.Query
   @type t :: %__MODULE__{}
 
-  alias Cambiatus.{Accounts.User, Commune.Community, Repo}
+  alias Cambiatus.{Accounts.User, Commune, Repo}
+  alias Cambiatus.Commune.Community
   alias Cambiatus.Shop.{Order, Product, ProductImage}
 
   schema "products" do
@@ -50,6 +51,8 @@ defmodule Cambiatus.Shop.Product do
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> cast_assoc(:images)
     |> validate_required(@required_fields)
+    |> validate_community_shop_enabled()
+    |> validate_track_stock_units()
   end
 
   def create_changeset(attrs) do
@@ -59,6 +62,44 @@ defmodule Cambiatus.Shop.Product do
     |> put_assoc(:community, attrs.community)
     |> foreign_key_constraint(:creator_id)
     |> foreign_key_constraint(:community_id)
+    |> validate_community_shop_enabled()
+    |> validate_track_stock_units()
+    |> validate_price_precision()
+  end
+
+  def validate_community_shop_enabled(changeset) do
+    community_id = get_field(changeset, :community_id)
+
+    Commune.get_community(community_id)
+    |> case do
+      {:ok, community} ->
+        if Map.get(community, :has_shop),
+          do: changeset,
+          else: add_error(changeset, :community_id, "news is not enabled")
+
+      {:error, _} ->
+        add_error(changeset, :community_id, "does not exist")
+    end
+  end
+
+  def validate_track_stock_units(changeset) do
+    unless get_field(changeset, :track_stock) do
+      add_error(changeset, :units, "cannot be filled if track_stock is false")
+    else
+      changeset
+    end
+  end
+
+  def validate_token_precision(changeset) do
+    community_id = get_field(changeset, :community_id)
+    price = get_field(changeset, :price)
+    {precision, _} = Eos.parse_symbol(community_id)
+
+    unless Float.round(price, precision) == price do
+      add_error(changeset, :price, "Price must have precision of #{precision}")
+    else
+      changeset
+    end
   end
 
   def from_community(query \\ Product, community_id) do
--- a/lib/cambiatus/shop/product.ex
+++ b/lib/cambiatus/shop/product.ex
@@ -38,8 +38,9 @@ defmodule Cambiatus.Shop.Product do
 
   # TODO: Put back this after the update of structure from blockchain to graphql
   # @required_fields ~w(title description price track_stock units created_block is_deleted)a
-  @required_fields ~w(title description price track_stock units created_block is_deleted created_tx created_eos_account created_at inserted_at updated_at)a
-  @optional_fields ~w(deleted_at image)a
+  @required_fields ~w(title description price track_stock community_id)a
+  @optional_fields ~w(units deleted_at image created_block is_deleted creator_id
+  created_tx created_eos_account created_at inserted_at updated_at)a
 
   @doc """
   This function contains the logic required for the validation of base shop changeset
@@ -64,13 +65,12 @@ defmodule Cambiatus.Shop.Product do
     |> foreign_key_constraint(:community_id)
     |> validate_community_shop_enabled()
     |> validate_track_stock_units()
-    |> validate_price_precision()
   end
 
   def validate_community_shop_enabled(changeset) do
-    community_id = get_field(changeset, :community_id)
-
-    Commune.get_community(community_id)
+    changeset
+    |> get_field(:community_id)
+    |> Commune.get_community()
     |> case do
       {:ok, community} ->
         if Map.get(community, :has_shop),
@@ -83,20 +83,12 @@ defmodule Cambiatus.Shop.Product do
   end
 
   def validate_track_stock_units(changeset) do
-    unless get_field(changeset, :track_stock) do
+    track_stock = get_field(changeset, :track_stock)
+    units = get_field(changeset, :units)
+
+    if (is_nil(track_stock) or track_stock == false) and
+         (not is_nil(units) and units > 0) do
       add_error(changeset, :units, "cannot be filled if track_stock is false")
-    else
-      changeset
-    end
-  end
-
-  def validate_token_precision(changeset) do
-    community_id = get_field(changeset, :community_id)
-    price = get_field(changeset, :price)
-    {precision, _} = Eos.parse_symbol(community_id)
-
-    unless Float.round(price, precision) == price do
-      add_error(changeset, :price, "Price must have precision of #{precision}")
     else
       changeset
     end
--- a/lib/cambiatus/shop/product.ex
+++ b/lib/cambiatus/shop/product.ex
@@ -35,7 +35,9 @@ defmodule Cambiatus.Shop.Product do
     has_many(:orders, Order, foreign_key: :product_id)
   end
 
-  @required_fields ~w(title description price track_stock units created_block is_deleted created_tx created_eos_account created_at)a
+  # TODO: Put back this after the update of structure from blockchain to graphql
+  # @required_fields ~w(title description price track_stock units created_block is_deleted)a
+  @required_fields ~w(title description price track_stock units created_block is_deleted created_tx created_eos_account created_at inserted_at updated_at)a
   @optional_fields ~w(deleted_at image)a
 
   @doc """
--- a/lib/cambiatus/shop/product.ex
+++ b/lib/cambiatus/shop/product.ex
@@ -45,26 +45,33 @@ defmodule Cambiatus.Shop.Product do
   @doc """
   This function contains the logic required for the validation of base shop changeset
   """
-  @spec changeset(Product.t(), map()) :: Ecto.Changeset.t()
-  def changeset(%Product{} = shop, attrs) do
-    shop
+  @spec changeset(Product.t(), map(), atom()) :: Ecto.Changeset.t()
+  def changeset(product, attrs, operation \\ :create)
+
+  def changeset(%Product{}, attrs, :create) do
+    %Product{}
     |> Repo.preload(:images)
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> cast_assoc(:images)
     |> validate_required(@required_fields)
-    |> validate_community_shop_enabled()
-    |> validate_track_stock_units()
-  end
-
-  def create_changeset(attrs) do
-    %Product{}
-    |> changeset(attrs)
-    |> put_assoc(:creator, attrs.creator)
-    |> put_assoc(:community, attrs.community)
     |> foreign_key_constraint(:creator_id)
     |> foreign_key_constraint(:community_id)
     |> validate_community_shop_enabled()
     |> validate_track_stock_units()
+  end
+
+  def changeset(%Product{} = product, attrs, :update) do
+    product
+    |> Repo.preload(:images)
+    |> cast(attrs, @required_fields ++ @optional_fields)
+    |> cast_assoc(:images)
+    |> validate_community_shop_enabled()
+    |> validate_track_stock_units()
+  end
+
+  def changeset(%Product{} = product, _, :delete) do
+    product
+    |> cast(%{deleted_at: DateTime.utc_now(), is_deleted: true}, [:deleted_at, :is_deleted])
   end
 
   def validate_community_shop_enabled(changeset) do
--- a/lib/cambiatus/shop/product.ex
+++ b/lib/cambiatus/shop/product.ex
@@ -111,7 +111,7 @@ defmodule Cambiatus.Shop.Product do
     units = get_field(changeset, :units)
 
     if (is_nil(track_stock) or track_stock == false) and
-         (not is_nil(units) and units > 0) do
+         (not is_nil(units) or units > 0) do
       add_error(changeset, :units, "cannot be filled if track_stock is false")
     else
       changeset
--- a/lib/cambiatus/shop/product.ex
+++ b/lib/cambiatus/shop/product.ex
@@ -111,7 +111,7 @@ defmodule Cambiatus.Shop.Product do
     units = get_field(changeset, :units)
 
     if (is_nil(track_stock) or track_stock == false) and
-         (not is_nil(units) or units > 0) do
+         (not is_nil(units) and units > 0) do
       add_error(changeset, :units, "cannot be filled if track_stock is false")
     else
       changeset
--- a/lib/cambiatus/shop/product.ex
+++ b/lib/cambiatus/shop/product.ex
@@ -32,14 +32,18 @@ defmodule Cambiatus.Shop.Product do
     belongs_to(:creator, User, references: :account, type: :string)
     belongs_to(:community, Community, references: :symbol, type: :string)
 
-    has_many(:images, ProductImage)
+    has_many(:images, ProductImage,
+      on_replace: :delete,
+      on_delete: :delete_all
+    )
+
     has_many(:orders, Order, foreign_key: :product_id)
   end
 
   # TODO: Put back this after the update of structure from blockchain to graphql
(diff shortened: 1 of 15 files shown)


lucca65:
@NeoVier com esse PR, vamos poder começar a usar o GraphQL para criar e editar anúncios. Vamos começar por esse passo primeiro, depois começamos a colocar dados novos, como categorias, disponibilidade de produtos, etc

lib/cambiatus/shop/product.ex
```diff
@@ -52,6 +63,35 @@ defmodule Cambiatus.Shop.Product do
...
+
+  def validate_community_shop_enabled(changeset) do
+    changeset
+    |> get_field(:community_id)
+    |> Commune.get_community()
+    |> case do
+      {:ok, community} ->
+        if Map.get(community, :has_shop),
+          do: changeset,
+          else: add_error(changeset, :community_id, "news is not enabled")
```
NeoVier, line 78: Should this be "shop is not enabled"?
```suggestion
          else: add_error(changeset, :community_id, "shop is not enabled")
```

lib/cambiatus/shop/product.ex
```diff
@@ -29,20 +32,28 @@ defmodule Cambiatus.Shop.Product do
...
     belongs_to(:community, Community, references: :symbol, type: :string)
 
+    has_many(:images, ProductImage)
     has_many(:orders, Order, foreign_key: :product_id)
   end
 
-  @required_fields ~w(title description price image track_stock units
-  created_block is_deleted deleted_at created_tx created_eos_account created_at)a
+  # TODO: Put back this after the update of structure from blockchain to graphql
+  # @required_fields ~w(title description price track_stock units created_block is_deleted)a
```
NeoVier, line 40: I'm not sure I understand this comment. Should it still be here?

lib/cambiatus/shop/product.ex
```diff
@@ -29,20 +32,28 @@ defmodule Cambiatus.Shop.Product do
...
+  @required_fields ~w(title description price track_stock community_id)a
+  @optional_fields ~w(units deleted_at image created_block is_deleted creator_id
+  created_tx created_eos_account created_at inserted_at updated_at)a
```
NeoVier, line 43: Does it make sense to have both `created_eos_account` and `creator_id`? What are the differences? Same thing for `created_at` and `inserted_at`

Also, most of the optional fields seem like they should be required:
- `created_block`
- `creator_id`
- `created_tx`
- `created_eos_account`
- `created_at`
- `inserted_at`
- `updated_at`

Is there a scenario where we don't have these? I guess the blockchain-related ones could be optional (until the transaction to create the order is made), but I'm not sure about the others

lib/cambiatus_web/resolvers/shop.ex
```diff
@@ -5,6 +5,36 @@ defmodule CambiatusWeb.Resolvers.Shop do
...
+  end
+
+  def upsert_product(_, params, %{context: %{current_user: current_user}}) do
+    params
+    |> Map.merge(%{creator_id: current_user.account})
+    |> Shop.create_product()
+    |> case do
+      {:error, reason} ->
+        Sentry.capture_message("Product update failed", extra: %{error: reason})
+        {:error, message: "Product update failed", details: Cambiatus.Error.from(reason)}
```
NeoVier, line 31:
```suggestion
        Sentry.capture_message("Product creation failed", extra: %{error: reason})
        {:error, message: "Product creation failed", details: Cambiatus.Error.from(reason)}
```

lib/cambiatus_web/resolvers/shop.ex
```diff
@@ -5,6 +5,36 @@ defmodule CambiatusWeb.Resolvers.Shop do
...
+
+      {:ok, product} = result ->
```
NeoVier, line 33: There's a warning here that `product` is unused
```suggestion
      {:ok, _product} = result ->
```

lib/cambiatus_web/schema/shop_types.ex
```diff
@@ -40,6 +40,20 @@ defmodule CambiatusWeb.Schema.ShopTypes do
...
+      arg(:community_id, non_null(:string))
+      arg(:title, non_null(:string))
+      arg(:description, non_null(:string))
+      arg(:price, non_null(:float))
+      arg(:images, non_null(list_of(non_null(:string))))
+      arg(:track_stock, non_null(:boolean))
+      arg(:units, :integer)
+
+      # TODO: Put authentication back
+      # middleware(Middleware.Authenticate)
```
NeoVier, line 54: Should we do this now or should it remain commented out?

lib/cambiatus_web/schema/shop_types.ex
```diff
@@ -56,15 +70,21 @@ defmodule CambiatusWeb.Schema.ShopTypes do
     field(:title, non_null(:string))
     field(:description, non_null(:string))
     field(:price, non_null(:float))
-    field(:image, :string)
     field(:track_stock, non_null(:boolean))
-    field(:units, non_null(:integer))
+    field(:units, :integer)
```
NeoVier, line 74: I wonder if we should have `track_stock` here... It creates the case where `track_stock` is true, but `units` is null. We can already know the value of `track_stock` just by checking if `units` is null or not.

lib/cambiatus_web/schema/shop_types.ex
```diff
@@ -56,15 +70,21 @@ defmodule CambiatusWeb.Schema.ShopTypes do
...
+
+    field(:inserted_at, non_null(:naive_datetime))
+    field(:updated_at, non_null(:naive_datetime))
+
+    # TODO: Remove deprecated fields
+    field(:image, :string, deprecate: true)
+    field(:created_block, non_null(:integer), deprecate: true)
+    field(:created_tx, non_null(:string), deprecate: true)
+    field(:created_eos_account, non_null(:string), deprecate: true)
+    field(:created_at, non_null(:datetime), deprecate: true)
```
NeoVier, line 87: Should we still keep these for now?

priv/repo/migrations/20220405214735_multiple_shop_images.exs
```diff
@@ -0,0 +1,23 @@
...
+
+  def change do
+    alter table(:products) do
+      add(:inserted_at, :naive_datetime, null: true)
+      add(:updated_at, :naive_datetime, null: true)
+    end
+
+    create table(:product_images) do
+      add(:product_id, references(:products))
+      add(:uri, :string, null: false, comment: "URI for the image or video")
```
NeoVier, line 12: Ooh, having videos for products is interesting!!

test/cambiatus_web/schema/resolvers/shop_test.exs
```diff
@@ -0,0 +1,116 @@
...
+        mutation {
+          product(id: #{product.id},
+                  description: "Nova descrição") {
+                    title
+                    description
+                  }
+        }
+      """
+
+      response =
```
NeoVier, line 101: There's a warning here that `response` is unused
```suggestion
      _response =
```

NeoVier (changes requested):
There's a bunch of comments just asking questions or reminding TODO comments.

I think we should at least address the `Product update failed` -> `Product creation failed` and `news is not enabled` -> `shop is not enabled` changes, as those could really trip us up in the future

lib/cambiatus/shop/product.ex
lucca65: we will have to wait for it to be migrated on prod first. Since we do the migration of data from the blockchain timestamps to regular timestamps, they are actually optional.

after it's migrated we will have to come back here and change field requirements as also remove those columns as they will be obsolete

lib/cambiatus/shop/product.ex
lucca65: those blockchain related will be all removed after we are done with data migration on prod.

the other ones we don't have to worry about since ecto handle them. but we will need to change NULL rules so it's back to the regular default elixir model

lib/cambiatus/shop/product.ex
lucca65: nice catch! thanks bro!

lib/cambiatus_web/resolvers/shop.ex
lucca65: there is two upsert function matches, this one only handle update

lib/cambiatus_web/schema/shop_types.ex
lucca65: this validation is made on product model, since it can be empty only depending on the value of track_stock

lib/cambiatus_web/schema/shop_types.ex
lucca65: yes, we will need to migrate data before removing those fields

priv/repo/migrations/20220405214735_multiple_shop_images.exs
lucca65: and for us super simple! just a uri! 🤠

lib/cambiatus_web/resolvers/shop.ex
NeoVier: Isn't this the create one? It calls `Shop.create_product()`. Either way, both of the function matches have `Product update failed`:
image

lib/cambiatus_web/schema/shop_types.ex
NeoVier: Yeah, but we won't know that on the frontend when we fetch it... We'll only see that `trackStock` is a `Bool` and `units` is a `Maybe Int`.

We can catch this on the `selectionSet`, so it won't ever bleed out onto the actual app logic, so it's not a big problem anyway

lib/cambiatus_web/resolvers/shop.ex
lucca65: oh damn, sorry bro, you are right! thanks!

lib/cambiatus_web/schema/shop_types.ex
lucca65: but it makes sense in a data design point of view right? Track stock is mandatory, but units will only be on a few scenarios.

lib/cambiatus_web/schema/shop_types.ex
lucca65: should be removed, thanks bro!

lib/cambiatus_web/schema/shop_types.ex
NeoVier: It does, but ideally we'd have a custom type like

```elm
type StockTracking
    = NoTracking
    | TrackUnits { units : Int }
```

I'm not really sure if it's doable with elixir/SQL/GraphQL though. I think Elm has spoiled me 😅

test/cambiatus_web/schema/resolvers/shop_test.exs
```diff
@@ -0,0 +1,116 @@
...
+
+      res = post(conn, "/api/graph", query: mutation)
+
+      response = json_response(res, 200)
+
+      assert %{
+               "data" => %{"product" => nil},
+               "errors" => [
+                 %{
+                   "details" => %{"community_id" => ["news is not enabled"]},
```
NeoVier, line 77: Tests are breaking due to the change on this string
```suggestion
                   "details" => %{"community_id" => ["shop is not enabled"]},
```

lib/cambiatus_web/schema/shop_types.ex
lucca65: well, we could use [union types](https://graphql.org/learn/schema/#union-types), I'll fiddle around to implement it. We have no way to do this on postgres/ecto, as far as I know, but we can on graphql

NeoVier (approved):
Looks good now! Just needs a fix on a test

lib/cambiatus_web/schema/shop_types.ex
lucca65: Bro, this is actually good but doesn't seem like I can do this type of query with no fields inside:

```graphql
mutation {
  product(id: 1) {
    ... on NoTracking {}
    ... on TrackingUnits { units } 
  } 
}
```

Makes some-ish sense, union types were thought when we had a type that describes them and redirect to different entities, like we do on the `notification.payload`... Guess we will have to keep it like that, at least for now

For the future we will have different types of products, like `digital`, `downloadable`, `nft`, `trackable` (in a has delivery sense), `stockable`. That will be used to allow different behaviors. Maybe we can add units as a type. We can think in more depth about this in the future!

NeoVier: Ok, awesome!

lib/cambiatus_web/schema/shop_types.ex
```diff
@@ -40,6 +40,19 @@ defmodule CambiatusWeb.Schema.ShopTypes do
...
+      arg(:title, non_null(:string))
+      arg(:description, non_null(:string))
+      arg(:price, non_null(:float))
+      arg(:images, non_null(list_of(non_null(:string))))
+      arg(:track_stock, non_null(:boolean))
+      arg(:units, :integer)
+
+      middleware(Middleware.Authenticate)
+      resolve(&Shop.upsert_product/3)
+    end
```
NeoVier, line 55: Should this have an optional `id` field? Otherwise, how do we update existing products?

Also, is `community_id` needed here? Can't we get the community through the absinthe context?

NeoVier:
Besides creating and updating, we also have `deletesale` on the smart contracts, which is missing in this PR. Are we treating the deletion through GraphQL as well?

lib/cambiatus_web/schema/shop_types.ex
lucca65: it does! I'll ping you in the test so you can have an example

test/cambiatus_web/schema/resolvers/shop_test.exs
```diff
@@ -0,0 +1,116 @@
...
+
+    test "update existing product" do
+      user = insert(:user)
+      product = insert(:product)
+
+      conn = build_conn() |> auth_user(user)
+
+      mutation = """
+        mutation {
+          product(id: #{product.id},
```
lucca65, line 93: @NeoVier check this test out, it contains an update mutation for reference, just send the ID

lib/cambiatus_web/schema/shop_types.ex
lucca65: About the community ID, its not on the absinthe context, although it could be. See, we use the `Community-Domain` header only on sign up as this is required for us to know how to handle invitations and netlinks (both happen on signIn and signUp).

We could add this as an automatic context for sure. To handle that I've opened this issue: https://github.com/cambiatus/backend/issues/223

test/cambiatus_web/schema/resolvers/shop_test.exs
NeoVier: That's weird bro, it doesn't show up in the schema (elm-graphql needs it to be on the schema so it can generate code), and it won't allow me to give it an id even if I try:

image

Also, all of the fields are required... How is that test passing? 🤔

lucca65: wow, this is weird... I'll check it out bro! oh, I'll implement a delete mutation as well

lucca65:
@NeoVier turns out I've done the update tests with `=` instead of `==` 🤦 

I've done improvements and implemented the `deleteProduct` mutation, let me know if it works properly for you!

NeoVier:
Welp, that must've been a fun one to debug 🤣

I'll test them out and let you know!

lib/cambiatus/shop/product.ex
```diff
@@ -29,29 +32,73 @@ defmodule Cambiatus.Shop.Product do
...
+        add_error(changeset, :community_id, "does not exist")
+    end
+  end
+
+  def validate_track_stock_units(changeset) do
+    track_stock = get_field(changeset, :track_stock)
+    units = get_field(changeset, :units)
+
+    if (is_nil(track_stock) or track_stock == false) and
+         (not is_nil(units) and units > 0) do
```
NeoVier, line 97: I think there's a logic error here... If I give it `track_stock = false` and don't provide `units`, it still says `cannot be filled if track_stock is false`.

Maybe we want `not is_nil(units) or units > 0`? Maybe even just `not is_nil(units)` is enough?

Here's the mutation I'm running:
```graphql
mutation {
  product(id: 57, trackStock: false) {
    trackStock
    units
  }
}
```

NeoVier (changes requested):

NeoVier:
Also, I can't add images to a product. When I try, I just get a 500 Something went wrong

Here's the mutation I'm trying:

```graphql
mutation {
  product(id: 57, images: ["https://cambiatus-uploads.s3.amazonaws.com/cambiatus-uploads/7c283b3afe8d47ffa7345c6dbbd865d2"]) {
    images
  }
}
```

lib/cambiatus/shop/product.ex
lucca65: ok, I've looked into this and what happens is that you are setting track_stock as false on a product that have units already. Do it like this and it will work:

```elixir
mutation {
  product(id: 57, trackStock: false, units: 0) {
    trackStock
    units
  }
}
```

NeoVier: Ok, that makes sense 🤦

NeoVier:
Uploading images works, but querying for the `images` field on old products gives a `Cannot return null for non-nullable field`. I see there's a migration to migrate that data, but when I try to run migrations, I just get `Migrations already up`:

image

Here's the mutation I'm trying, and the error I'm getting:

```graphql
query {
  products(communityId: "0,BUSS") {
    images {
      uri
    }
  }
}
```

image

lucca65:
Hours of debugging, database dumping for a single line fix ✊😮‍💨

NeoVier (approved):
It's working now! If you want, you can do your own testing on the frontend PR here: https://github.com/cambiatus/frontend/pull/734

The only thing that I'm not sure is intended is that the old `image` is not included in the new `images` list. It's ok for now from the frontend point of view, but might be a little more annoying once we start working on supporting multiple images

NeoVier:
Oh, I just realized: is this PR problematic for `ProductPreview`s? Are we supporting multiple images on those as well? Are they consuming the new `images` that we now upload?

lucca65:
> The only thing that I'm not sure is intended is that the old image is not included in the new images list. It's ok for now from the frontend point of view, but might be a little more annoying once we start working on supporting multiple images

Bro no need to worry too much about this, we will have to do a migration in such a way that we will not need to handle this. You can always simply show `Images` and drop checks or usage on `image`

NeoVier (approved):
Images are working 📷 👍

PR #227, opened 2022-05-08
Created rich_link_controller and functionalities

### LINKED ISSUE #182: Add rich link preview to the shop
When users share Cambiatus links it is shown plain, without a preview

Implement a preview with the product image and the community logo. Maybe with some text like "${user} is selling ${product} on Cambiatus, check it out"

N/A

**Additional context**
N/A

**Screen shots**
Captura de Tela 2021-05-12 às 12 39 48

MatheusBuss:
Closes #182

## Changes Proposed (a list of new changes introduced by this PR)
Add rich link/preview when Cambiatus andand associated links are shared in different social medias

1. Run `mix phx.server`
2. Send HTTP request for differente funcionalities
  2.1. For User sharing send `GET` request to `/api/rich_link/profile/#{user_name}`
  2.2. For Product sharing send `GET` request to `/api/rich_link/shop?id=#{product_id}`
  2.3. For Community sharing send `GET` request to `/api/rich_link` (Use the host header in the HTTP request for the appropriate community)
3. Check html info sent by the server

Example of sharing html generated by these changes through Whatsapp:

image


MatheusBuss:
The Elixir part should be working now. Sharing the HTML generated by the phoenix server worked on WhatsApp, Slack and Facebook. Twitter was not yet tested.

Some of the next steps that I have discussed with @NeoVier:
1 - Use NGINX to redirect the `GET` request from social media web crawlers based on their user agents (we can look into this list https://developers.whatismybrowser.com/useragents/explore/software_name/)
2 - Check if all the fields are beign filled correctly according to the Open Graph Protocol (https://ogp.me/)

lib/cambiatus_web/router.ex
```diff
@@ -51,6 +51,7 @@ defmodule CambiatusWeb.Router do
     get("/chain/info", ChainController, :info)
     post("/invite", InviteController, :invite)
     get("/manifest", ManifestController, :manifest)
+    get("/rich_link/*page", RichLinkController, :rich_link)
```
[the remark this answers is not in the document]
NeoVier: At the end of the day, we could use any url we want, since we will rewrite urls with nginx, right?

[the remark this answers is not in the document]
lucca65: right, the prefix don't really matter, but I'm suggesting that we drop `/profile` and `/shop` and infer it based on the data type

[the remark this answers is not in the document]
MatheusBuss: We can certainly do it this way too. I was just thinking that it would be easier to implement the current iteration on NGINX, since ot only requires to add `/api/rich_link/` to the requested path.

But we could look into doing the logic of tranforming the URL into the way you mentioned.

MatheusBuss:
> bro, when the bio has markdown we don't render it properly, like you can see on https://api.staging.cambiatus.io/api/rich_link/profile/lucca (I've published it to staging).

Using [`pandex`](https://github.com/FilterKaapi/pandex) I was able to convert markdown to plain text, so it shouldn't be a problem then. But this alternative requires the installation of `pandex` itself and [Pandoc](https://pandoc.org/).

I've tried approaching it using [`Earmark`](https://hexdocs.pm/earmark/Earmark.html) (which is already on `mix.exs`), but it only converts into HTML os AST as far as I can tell. Therefore we eould need to convert this HTML into plain text too.

So I believe we should try to use `pandex` going forward

lib/cambiatus_web/router.ex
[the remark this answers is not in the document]
lucca65: I guess you could do the data type checking on nginx as well, no problem for me. Maybe we can have the NGINX configurations in our README setup section, so this don't get lost if we switch servers in the future.

lucca65:
Ok @MatheusBuss! Np for pandoc

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -23,11 +24,24 @@ defmodule CambiatusWeb.RichLinkController do
 
     case data do
       {:ok, data} ->
-        response = EEx.eval_file(@rich_link_file, Enum.into(data, []))
-        html(conn, response)
+        response =
+          data
+          |> Enum.map(fn {k, v} -> {k, md_to_txt(v)} end)
+          |> Enum.into([])
```
lucca65, line 30: Since we only allow markdown on description (both for profile and product) shouldn't we only try to parse for those fields? The way it is we are putting every field to possible parsing. It will be resource heavy to do this for every field on every request when we don't need to

lucca65 (changes requested):
Also tests? 😬 🔧

lib/cambiatus_web/controllers/rich_link_controller.ex
MatheusBuss: Sure thing, I'll be doing that then

MatheusBuss: Done!

MatheusBuss:
Still to do:
- Use Phoenix's views to render the template
- Implement NGINX redirect
- Rewrite some template fields to properly comply to the Open Graph Protocol

lucca65 (changes requested):
Its getting better and better bro! Keep it up!

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -3,13 +3,12 @@ defmodule CambiatusWeb.RichLinkController do
 
   @moduledoc """
   Get data and render html to be used for rich links (also known as Open Graphs).
-  These rich links show additional information about the website when shared on social media and must be compliant with the [Open Grap Protocol](https://ogp.me/)
+  These rich links show additional information about the website when shared on social media
+  and must be compliant with the [Open Grap Protocol](https://ogp.me/)
   """
```
lucca65, line 8: please move `@moduledoc` to the first statement on the module, instead of the `use` macro

lucca65 (changes requested):

lucca65:
a URL precisa ser pra tela do perfil, ao invés do email

README.md
```diff
@@ -120,6 +124,25 @@ Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
...
+
+- Note 2: For Windows users it is recommended to install NGINX under the C:\Program Files directory
+
+**Step 2**
+
+Open a terminal and `cd` into NGINX's installation directory and run
+```
+nginx
+```
+Now you can make HTTP requests to NGINX on your browser under localhost!
```
lucca65, line 145: Maybe we can simply link to nginx documentation, the part that we need is only the relevant configuration changes

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -0,0 +1,93 @@
...
+      {:error, reason} ->
+        send_resp(conn, 404, reason)
+    end
+  end
+
+  def product_rich_link(id, community_subdomain) do
+    case Shop.get_product(nil, %{id: id}, nil) do
+      {:ok, product} ->
+        product = Repo.preload(product, :images)
+        [image | _] = product.images
```
lucca65, line 44: This may fail, images can be an empty array. Maybe this is a good way to explore https://www.openmymind.net/Elixirs-With-Statement/

lib/cambiatus_web/views/rich_link_view.ex
```diff
@@ -0,0 +1,16 @@
+defmodule CambiatusWeb.RichLinkView do
+  use CambiatusWeb, :view
+
+  require Earmark
+  require HtmlSanitizeEx
+
+  def md_to_txt(markdown) do
+    with {:ok, string, _} <- Earmark.as_html(markdown) do
+      HtmlSanitizeEx.strip_tags(string)
+      |> String.trim()
```
lucca65, line 10: if there is a pipe, use it all the way:

```suggestion
      string
      |> HtmlSanitizeEx.strip_tags()
      |> String.trim()
```

nginx/nginx.conf
```diff
@@ -0,0 +1,217 @@
...
+	server {
+			listen 80;
+			listen [::]:80;
+
+			server_name staging.cambiatus.io *.staging.cambiatus.io;
+
+			return 301 https://$host$request_uri;
+			return 404; # managed by Certbot
+	}
+}
```
lucca65, line 217: This file is huge, no way we need 207 lines to describe this. You can add only the relevant part and it can be directly under our README.md file

test/cambiatus_web/controllers/rich_link_controller_test.exs
```diff
@@ -0,0 +1,124 @@
...
+      # Submit GET request for a user rich link
+
+      conn =
+        %{conn | host: community.subdomain.name}
+        |> get("/api/rich_link/profile/#{user.account}")
+
+      response = html_response(conn, 200)
+
+      # Check http code 200 and if all the rich link fields are properly filled
+      assert conn.status == 200
```
lucca65, line 69: +1

lucca65 (changes requested):
Sorry for the long time to work on the review @MatheusBuss

lucca65 (changes requested):
I liked the instructions. Just followed it now and tested on telegram.

here is the result for this URL: https://api.staging.cambiatus.io/api/rich_link/profile/lucca
Captura de Tela 2022-05-24 às 11 54 28

I'm missing the photo and we also have a <u> tag that was not rendered correctly. 

Also its rendering some sort of lorem ipsum, even if my profile don't have it:

Captura de Tela 2022-05-24 às 11 57 46

MatheusBuss:
> Also its rendering some sort of lorem ipsum, even if my profile don't have it

That's odd. I tested on Telegram and had the same result, but on WhatsApp it renders correctly (other than the underline tag and the image)
image

Even checking the html on the page it doesn't have any lorem ipsum in it. Don't know why it shows on Telegram, maybe it's cached from previous tests?

> I liked the instructions. Just followed it now and tested on telegram.

We should try to use an URL withouth the `/api/rich_link` to test if nginx is redirecting correctly the crawlers (something like [https://buss.staging.cambiatus.io/profile/mattheusbuss](https://buss.staging.cambiatus.io/profile/mattheusbuss)). I even tried with this link, but it didn't work.

MatheusBuss:
> Don't know why it shows on Telegram, maybe it's cached from previous tests?

Exactly this. I found [@webpagebot](https://telegram.me/webpagebot) on [this thread](https://stackoverflow.com/questions/34707915/how-do-you-clear-the-open-graph-cache-of-an-url-on-telegram) . It clears the cache on telegram. Now the description is updated!

image

lucca65:
Great bro this is awesome! Can we also preview the profile image when sharing profile links?

MatheusBuss:
I was looking into it, and the html does include the image link, but it isn't loading. I tried downloading based on the link and noticed that the file doesn't have the extension (.jpeg or .png) so it could be that.
Another reason might be the image size. Facebook limits the image size to 8MB ([per the developer page](https://developers.facebook.com/docs/sharing/webmasters/images/)) but I remember reading some recommendations to limi the size to 30 kB, but I will have to double check on that.

MatheusBuss:
> I see bro, in this case can you open a new issue for us to transform images on this link? This way we can tackle this in a part two

Created #239

lucca65 (approved):
Looks good to me! congrats bro!

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def product_rich_link(id, community_subdomain) do
+    get_image = fn product ->
+      with product = Repo.preload(product, :images),
+           [image | _] <- product.images do
+        image.uri
+      else
+        _ ->
+          "https://buss.staging.cambiatus.io/images/logo-cambiatus-mobile.svg"
```
NeoVier, line 44: I don't think it's a good idea to point to a staging community to get this logo. If someday we decide to change the name of that file in staging, it could affect production. Maybe @lucca65 can give better directions on where to store this (on the server itself, on AWS, or somewhere else), but here's the image for you to use:

logo-cambiatus-mobile

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -0,0 +1,97 @@
...
+      end
+    end
+
+    case Shop.get_product(nil, %{id: id}, nil) do
+      {:ok, product} ->
+        {:ok,
+         %{
+           description: product.description,
+           title: product.title,
+           url: community_subdomain,
```
NeoVier, line 54: From https://ogp.me:

> og:url - The canonical URL of your object that will be used as its permanent ID in the graph, e.g., "https://www.imdb.com/title/tt0117500/".

I think we should use the product's url here, which would look something like `https://buss.staging.cambiatus.io/shop/#{product.id}`

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def user_rich_link(account, community_subdomain) do
+    case Accounts.get_user(nil, %{account: account}, nil) do
+      {:ok, user} ->
+        {:ok,
+         %{
+           description: user.bio,
+           title: user.name,
+           url: community_subdomain,
```
NeoVier, line 71: Same thing as the product url. The profile url looks like `https://buss.staging.cambiatus.io/profile/#{user.account}`

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -0,0 +1,97 @@
...
+    end
+  end
+
+  def user_rich_link(account, community_subdomain) do
+    case Accounts.get_user(nil, %{account: account}, nil) do
+      {:ok, user} ->
+        {:ok,
+         %{
+           description: user.bio,
+           title: user.name,
```
NeoVier, line 70: I'm not sure if this is true in the backend as well, but in the frontend, `user.name` can be null. Maybe we should check this here, and display `user.account` if `user.name` is null

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -0,0 +1,97 @@
...
+      with community_subdomain <- conn.host do
+        case Map.get(params, "page") do
+          ["shop", id] ->
+            product_rich_link(id, community_subdomain)
+
+          ["profile", account] ->
+            user_rich_link(account, community_subdomain)
+
+          _ ->
+            community_rich_link(community_subdomain)
```
NeoVier, line 24: Nice 🎉

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    case Shop.get_product(nil, %{id: id}, nil) do
+      {:ok, product} ->
+        {:ok,
+         %{
+           description: product.description,
+           title: product.title,
+           url: community_subdomain,
+           image: get_image.(product),
```
NeoVier, line 55: Elixir syntax is so weird 💀

NeoVier (changes requested):

lib/cambiatus_web/controllers/rich_link_controller.ex
MatheusBuss: We may have to look more into this, since it seems that [svg's are not accepted on OG's](https://developers.facebook.com/docs/sharing/webmasters/#images](https://indieweb.org/The-Open-Graph-protocol#Does_not_support_SVG_images)). We should look instead into [using, jpeg, png or even gif](https://developers.facebook.com/docs/sharing/webmasters/#images).

Maybe this is something that can be addressed on #239.

lib/cambiatus_web/controllers/rich_link_controller.ex
MatheusBuss: Thanks for the reviews! 😁

PR #237, opened 2022-05-18
Feature/shop categories

### LINKED ISSUE #224: Implementação de categorias na loja
Categories should have:

- [x] A parent community
- [x] Allow a single icon image to it
- [x] Allow a banner image to it
- [x] Products can belong to multiple categories (`product_category`: `product_id`, `category_id`, `position`. Unique indexes)
- [x] Categories will have a `slug` and can be previewed if its a bot that is searching for it
- [x] Meta fields: `meta-title`, `meta-description`, `meta-keywords`
- [x] have a markdown enabled description
- [x] Adjust event-source to automatically create those default values on create_community handler 👇 
- [x] Categories should allow for subcategories, in a self referencing structure. This should allow for a tree-like organization and nesting
- [x] Needed GraphQL mutations
- [x] Gerar slugs usando ID. se o slug vier como `art` gerar slug como `art--#{category_id}`
- [x] #242
- [x] #243
  - Art and craftsmanship
  - Self-care
  - Food and drinks
  - Footwear
  - clothing and accessories
  - Courses
  - consultancies and mentoring
  - Events
  - Books and Publicantions
  - Furniture and household items
  - Household services and repairs

lucca65:
Closes #224 
will interact with, but not finish #223

- [x] Move files around to respect Phoenix Style and to be compatible with generators like `mix phx.gen.context`
- [x] Add context operations for category
- [x] Add category and sub category models
- [x] Add product category relationship
- [x] Allow admin edit for products
- [x] Add GraphQL mutations for adding categories to an product
- [x] Add GraphQL mutation for creating, updating and deleting categories
- [x] Add tests for GraphQL mutations. Test out different permission setups
- [x] Make sure GraphQL mutations are allowed only by admin or product owner


lib/cambiatus/shop/product.ex
```diff
@@ -82,21 +90,44 @@ defmodule Cambiatus.Shop.Product do
...
+
+          _ ->
+            false
+        end
+      end)
+
+    if any_invalid? do
+      add_error(changeset, :product_category, "Can't find category with given ID")
+    else
+      changeset
```
NeoVier, line 125: Maybe a fold/reduce would be better here:

```elixir
def validate_categories(...) do
  Enum.reduce(product_categories, changeset, fn product_category ->
    with ... do
      changeset
    else
      add_error(...)
    end
  end)
end
```

NeoVier (commented):
Left a few comments, but it looks good! I'll test it better when I start working on the frontend part of this

Are we going to generate the slug automatically or is it defined by the admin?

lib/cambiatus_web/schema/shop_types.ex
```diff
@@ -51,10 +51,41 @@ defmodule CambiatusWeb.Schema.ShopTypes do
...
+
       middleware(Middleware.Authenticate)
       resolve(&Shop.upsert_product/3)
     end
 
+    @desc "[Auth required - Admin only] Upserts a category"
+    field :category, :category do
+      arg(:id, :integer)
+
+      arg(:category_id, :string, description: "Parent category ID")
```
NeoVier, line 66: Shouldn't this be an `int`?

lib/cambiatus_web/schema/shop_types.ex
```diff
@@ -114,6 +147,27 @@ defmodule CambiatusWeb.Schema.ShopTypes do
...
+    field(:image_uri, :string)
+    field(:name, non_null(:string))
+    field(:description, non_null(:string))
+
+    field(:slug, :string)
+    field(:meta_title, :string)
+    field(:meta_description, :string)
+    field(:meta_keywords, :string)
+
+    field(:category, :category, resolve: dataloader(Cambiatus.Shop))
```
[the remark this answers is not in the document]
lucca65: done!

lib/cambiatus/shop/product.ex
lucca65: wow, nice! I'm not used to use list iterators! Loved the end result!

NeoVier:
@lucca65 since we're going to allow users to reorder categories, would it be a good idea to have a mutation to do that?

We could do it with two `category` mutations (one to add the category to the new parent, one to remove it from the old parent), but if one of them fails, we might get duplicate data or even lose the category

NeoVier:
@lucca65 I'm not exactly sure how we're going to use the `keywords` field, but I think it makes more sense as a list of strings, what do you think?

NeoVier:
@lucca65 also, can we accept empty slugs? If we don't, we can't really create categories with (only) Amharic characters (which generate an empty slug)

lucca65:
> @lucca65 I'm not exactly sure how we're going to use the keywords field, but I think it makes more sense as a list of strings, what do you think?

Vamos usar na hora de renderizar os preview links que iremos implementar em #231 junto com outras melhorias e mudanças em produtos. As categorias serão usadas para gerar os slugs de cada produto. Posteriormente podemos ter slugs para gerar as keywords de cada categorias se elas tiverem páginas próprias

Faz sentido sim, no fim a gente vai mandar pros bots só uma string com valores separados por virgula. No front poderíamos usar o mesmo elemento que usamos hoje nos interesses do perfil.... Imagino que é pra isso que você perguntou né?

lucca65:
> @lucca65 also, can we accept empty slugs? If we don't, we can't really create categories with (only) Amharic characters (which generate an empty slug)

Nesse caso acho que precisamos exigir que seja preenchido com alguma coisa, se o slug for vazio vamos quebrar a estrutura de link. Como é uma limitação geral e não específica do nosso caso, acho que é tranquilo obrigarmos o usuário a preencher. 

Se for necessário podemos colocar um tooltip explicando mais

lucca65:
> É isso mesmo 😅. Se ficar mais fácil, podemos tratar como uma lista no front, e mandar como uma única string pro back

Sussa manda bala!

NeoVier:
@lucca65 it seems like the position isn't being persisted...

Here's an example mutation:

```graphql
mutation {
  category(id: 3, categories: [{id: 24, position: 0}, {id: 5, position: 1}]) {
    categories {
      id
      position
    }
  }
}
```

And here's what I get back from it:

```json
{
  "data": {
    "category": {
      "categories": [
        {
          "id": 24,
          "position": 0
        },
        {
          "id": 5,
          "position": 0
        }
      ]
    }
  }
}
```

Note that both of the categories have position `0`

NeoVier:
Also, the way it is now makes it hard to reorder root categories. Say I have categories 0, 1 and 2, all of which have no parent. If I want to rearrange them to the order 0, 2, 1, I need to create a mutation for 2 (to change its position to 1), and a mutation for 1 (to change its position to 2). If the list was larger, I'd have to do even more mutations.

I'm not sure what's a good approach for this:
1. Have another mutation just to reorder categories (with an optional `id` field - if no `id` is provided, it affects all of the root categories)
2. Do some work on the backend to shift categories based on the moved category (we only ever move one at a time)
3. Something else?

lucca65:
Yeah, that's one of the reasons I didn't want to implement ordering for now. Ordering lists is super easy, but trees not so much. In fact a lot of tree sorts implementations make their way into competitive programming challenge and to be honest that's not something I'm used to do... 

We also lack libraries, at least by my search on hex.pm 

It would be an interesting challenge for sure. For now I've fixed the ordering issue by creating a database transaction, updating every child ordering, then updating its parent relationship. 

To be honest bro, I think we need to invest more into tree/graphs explorations, even if the tools around are scarce. Maybe we can rework categories using graph capabilities that Postgres has or even exploring new database options. This might be fertile ground for our future work and some of our queries, like the ones @MatheusBuss is working on #252 would be way simpler if we didn't stuck to relational databases, as we would leverage those graph algorithms to those use cases.

In anyway, ideally we would pause this and implement it on the best way possible, but with a meeting for categories scheduled for tomorrow and several months with categories requested, I think we should hammer this issue with this incomplete and troublesome implementation. I wish we could avoid this entirely until we had an final optimized solution, because this tree use case is the first step to a lot more to come on Cambiatus. 

For now, lets stick to this basic implementation and if its giving you too much of a headache on the frontend, we could limit this functionality

lucca65:
ok! Deploy feito! Publiquei em staging também. Adicionei um caso de teste para essa query que você mandou!

NeoVier (approved):
Looks good! 💪

PR #246, opened 2022-06-08
Feature/categories filter for shop products

### LINKED ISSUE #242: Accept categories filter as a list on `products` query

MatheusBuss:
Closes #242 

- [x] Create filter for products based on categories
- [x] Create test for filter

Run `mix test test/cambiatus_web/schema/resolvers/shop_test.exs`


MatheusBuss:
This implementation is failing because the join results in duplicate entries when a product is in two categories being queried at the same time. This is still to be fixed.

NeoVier:
@MatheusBuss since this depends on #237 , it might be a good idea to change the target branch to be `feature/shop-categories`. If you don't, and you finish this before #237, you will end up pushing all of the code of that PR to `master` before it's done. It will also make this PR a lot easier to review, since it will only show the differences between this PR and #237 

(At least I think so, maybe @lucca65 can correct me if I'm wrong 😅)

To do it, click "Edit" on the top of the page, and change the dropdown that says "base:master" to the correct branch

Untitled

lucca65:
@NeoVier acho que podemos mirar a master aqui, mas com a disciplina de não mergear/aprovar até o outro PR, acho que assim fica melhor do que ter um PR grandão.

test/cambiatus_web/schema/resolvers/shop_test.exs
```diff
@@ -429,8 +492,7 @@ defmodule CambiatusWeb.Resolvers.ShopTest do
       response3 =
         conn
         |> post("/api/graph", query: query_3)
-        |> json_response(200)
```
lucca65, line 260: why remove this line?

lucca65 (changes requested):

test/cambiatus_web/schema/resolvers/shop_test.exs
MatheusBuss: Happened by accident when pulling the origin into the branch. Had a conflict with another PR.

lucca65 (approved):
✅ Awesome! I'll merge it right away!

Issue #198, opened 2021-11-22
Create Shop orders


lucca65:
**Is your feature request related to a problem? Please describe.**
Today we have products and a very basic structure called `orders`. Before it was called `sale_history` because it was really only a registration of buying something from the shop. This proposal suggests a new structure to make the `order` abstraction more powerful and more useful.

**Describe the solution you'd like**
First we need to make the order have status. Orders always relate to trying to buy products from the same seller, or user. An order looks like this:

- **Product ID list** which contains a list of the product IDs that the buyer wants
- **Quantity**
- **Amount**
- **Status** status of the order. This first draft will allow only the basic ones: `created`, `completed`
- **Payment** a JSON or even only a transaction ID. allow us to link an order with a transfer
- **Timestamps** for creation/updates

Another change is that today, orders are created by listening to [this contract call](https://github.com/cambiatus/contracts/blob/master/community/community.hpp#L234). We should change this to make the creation of an Order to be an GraphQL call. After the order is created we can update it with the transaction ID to link it with the payment. This process can later be expanded to include several payments, until the oder is completed.

With that we can completely remove the `transfer_sale` call and properly using the `order` abstraction.

We will need the following screens:

- My orders (as a seller)
- My orders (as a buyer)
- Community orders (for admins)
- Community stats of orders
- Individual order 
  - Notion of quantities for products
  - Notion of payments on its own area
  - Contact information
  - Status of the order (a way for the user to predictably know what's next -- wait for the seller, etc)

**The new features users may gain**

- This will allow us to create interfaces to both seller and buyer to know how their orders are. 
- It will be easy to add steps related to shipping or partial payments too
- This will also allow sellers to have greater control, with dedicated shop pages to search and see their orders
- This will make the timeline possible, with our current structure we cannot achieve the result shown in the designs

**Describe alternatives you've considered**
We could start working on the idea of expanding possible `status`es. We could choose to track partial payments (useful for payments with fiat and crypto). We could add tracking capabilities to allow users to inform delivering/posting information and also add stuff like `canceled` or other abstractions

**Additional context**

lucca65:
### Entity and abstractions

#### Product
- name
- slug (to be used on the url)
- price
- community_id
- creator
- available_on
- discontinue_on
- categories list_of(int)
- properties (key value list to be used freely by seller and displayed on product page)
- multiple images

#### Orders
- id (uuid)
- creator
- community
[...]

PR #258, opened 2022-07-13
Enhancement/email unsubscribing without password

### LINKED ISSUE #204: Allow unsubscribing from our emails without having to login in our frontend
# Story

I, as an community member
Want to be able to adjust my email communication preferences, even I lost my 12 words
To keep my sanity and privacy

### Telas

N/A

### Casos de uso

Given someone received any of our emails
When clicked "Unsubscribe" 
Then our user will be redirected to a simple unsubscribe page

Given any user accessed the unsubscribe page
When they choose a preference
Then its updated

Given that the user will not on the frontend app while doing this
When accessing the preference page, they must be able to have a single usage access token
Then, after doing their changes, their access token is set to be revoked, so no one can reuse it

### Descrição do Problema

After implementing #195, we got an edge case where the user will not be able to unsubscribe if they lose their 12 words. Since this can lead us to not respect the users right to not receive any emails from us, we need to address this.

MatheusBuss:
Closes #204 

- [x] Add authenticated link to sent emails
  - [x] Generate token for user authentication
  - [x] Generate a token for each user
  - [x] Insert `list-unsubscribe` and `list-unsubscribe-post` headers according to [RFC 8058](https://datatracker.ietf.org/doc/html/rfc8058)
  - [x] Render link for cancelling subscription in all sent email
- [x] Authenticate user when using the provided link
  - [x] Ensure that the authentication is only valid for unsubscribing to emails
- [x] Implement one click functionality for unsubscription
- [x] Handle verification for a page specifically made for unsubscribing


lib/cambiatus_web/auth_token.ex
```diff
@@ -4,16 +4,38 @@ defmodule CambiatusWeb.AuthToken do
...
   end
 
-  def auth_salt() do
+  def auth_salt_user() do
     Application.get_env(:cambiatus, :auth_salt)
   end
+
+  def auth_salt_email() do
+    Application.get_env(:cambiatus, :auth_salt_email)
+  end
```
lucca65, line 40: Maybe you can match on the function itself

```elixir
def auth_salt("user"), do: Application.get_env(:cambiatus, :auth_salt)
def auth_salt("email"), do: Application.get_env(:cambiatus, :auth_salt_email)
```

lib/cambiatus_web/controllers/unsubscribe_controller.ex
```diff
@@ -0,0 +1,18 @@
...
+  use CambiatusWeb, :controller
+
+  alias CambiatusWeb.AuthToken
+
+  def one_click(conn, %{"token" => token, "subject" => subject} = _params) do
+    with {:ok, %{id: account}} <- AuthToken.verify(token, "email"),
+         %Cambiatus.Accounts.User{} = current_user <- Cambiatus.Accounts.get_user(account) do
+      CambiatusWeb.Resolvers.Accounts.update_preferences(nil, %{subject => false}, %{
+        context: %{current_user: current_user}
+      })
```
lucca65, line 13: It doesn't make sense for Resolvers and Controllers to be talking. They are from different pipelines of how we handle an Request. Resolvers will be used by GraphQL/Absinthe to sort our their own scope.

All other requests should be handled in they own way. Not only this decouples our logic for each API type (REST has one structure and GraphQL has another. If we implement RTC it would use another) but it also gives us hints on our own architecture. 

If we find ourselves having to share logic between web adapters, we surely are having business rules inside our web layer. 

We should try to keep them separate as it will help us grow more easily. Necessary bureaucracy. Hit me up on slack if you need to talk more about this bro!

lib/cambiatus_web/auth_token.ex
MatheusBuss: Definitively way better

lib/cambiatus_web/controllers/unsubscribe_controller.ex
MatheusBuss: Yeah, it makes sense. I wasn't too sure of how to implement this. It made sense to re utilize what we already had, but the way I made really didn't feel right. I guess I'll be doing it via changeset then.

lucca65: Well, I can see two scenarios:

1. We do change sets inside the resolver. If we do this, that's wrong, we should send all the necessary data to our `lib/cambiatus/` modules to do so
2. We already update preferences in our business layer. In that case you can use the function

Ideally we are able to hook up any existing feature into our business layer even if we are rebuilding the API from scratch

lib/cambiatus_web/controllers/unsubscribe_controller.ex
```diff
@@ -1,18 +1,16 @@
...
+  @moduledoc false
+
+  use CambiatusWeb, :controller
+
+  alias CambiatusWeb.AuthToken
+
+  def one_click(conn, %{"token" => token, "subject" => subject} = _params) do
+    with {:ok, %{id: account}} <- AuthToken.verify(token, "email"),
+         %Cambiatus.Accounts.User{} = current_user <- Cambiatus.Accounts.get_user(account) do
+      Cambiatus.Accounts.update_user(current_user, %{subject => false})
```
lucca65, line 11: Maybe an alias would make this easier to read?

MatheusBuss:
Acredito que é mais ou menos isso mesmo, mas tem algumas coisas que ainda poderiam ser feitas:
* Adicionar um teste para o corpo html do email
* Verificar se o one-click unsub está funcionando
------
Outra coisa que vi durante o desenvolvimento desse PR que poderíamos fazer: O gmail possui o que ele chama de [actions](https://developers.google.com/gmail/markup/actions/actions-overview), que são botões ao dar hover no email que fazem alguma coisa, como o exemplo do sentry abaixo. Quem sabe poderia ser algo a fazer também.
image

lib/cambiatus_web/email.ex
```diff
@@ -21,55 +25,93 @@ defmodule CambiatusWeb.Email do
...
+    |> header("List-Unsubscribe", one_click_unsub(claimer, community, "claim_notification"))
+    |> header("List-Unsubscribe-Post", "List-Unsubscribe=One-Click")
+    |> render_body("claim.html", render_params(claim))
     |> Mailer.deliver()
   end
 
   # input is a community with preloaded news with less than 30 days and members with active digest
   def monthly_digest(community) do
     Enum.each(community.members, fn member ->
       new()
```
NeoVier, line 60: I wonder if this is going to work... Looking at the SES console, we have limits on how many emails we can send (the main worry is emails/second):
image

Maybe @lucca65's idea of sticking the emails into a queue would be nice to avoid getting rate limited

lib/cambiatus_web/auth_token.ex
```diff
@@ -4,16 +4,16 @@ defmodule CambiatusWeb.AuthToken do
...
-    Phoenix.Token.sign(Endpoint, auth_salt(), %{id: user.account})
+  def sign(user, authorization \\ "user") do
+    Phoenix.Token.sign(Endpoint, auth_salt(authorization), %{id: user.account})
   end
 
   @doc "Decodes original data from given `token` and verifies its integrity"
-  def verify(token) do
-    Phoenix.Token.verify(Endpoint, auth_salt(), token, max_age: 365 * 24 * 3_600)
+  def verify(token, authorization \\ "user") do
+    Phoenix.Token.verify(Endpoint, auth_salt(authorization), token, max_age: 365 * 24 * 3_600)
```
NeoVier, line 13: I remember we talked about this before, but I don't really remember what we settled on, so...

Should we worry about setting a lower `max_age`, or invalidating it after one use (or something like that) for email tokens? I don't think it's a huge security issue, but if there's someone inspecting the network traffic, they can get the user's token and act in their behalf

lib/cambiatus_web/schema/middleware/email_special_authenticate.ex
```diff
@@ -0,0 +1,23 @@
...
+      %{current_user: _} ->
+        resolution
+
+      %{user_unsub_email: _} ->
+        resolution
+
+      _ ->
+        resolution
+        |> Absinthe.Resolution.put_result(
+          {:error, "Please sign in firt or use a valid unsubscription link"}
```
NeoVier, line 19:
```suggestion
          {:error, "Please sign in first or use a valid unsubscription link"}
```

lib/cambiatus_web/views/email_view.ex
```diff
@@ -1,12 +1,3 @@
...
-
-  alias Earmark
-
-  @doc "Renders the transfer email memo from markdown to html"
-  def render("transfer.html", %{transfer: transfer}) do
-    {:ok, html, []} = Earmark.as_html(transfer.memo)
-
-    html
-  end
 end
```
NeoVier, line 3: Should we still keep this file if it's empty? Maybe I'm missing some elixir magic 😅

lib/cambiatus_web/email.ex
```diff
@@ -21,55 +25,93 @@ defmodule CambiatusWeb.Email do
...
+  defp render_params(%User{} = user, %Community{} = community) do
+    unsub_link = unsub_link(user, community, user.language)
+    %{community: community, user: user, unsub_link: unsub_link}
+  end
+
+  def one_click_unsub(member, community, subject) do
+    token = AuthToken.sign(member, "email")
+
+    "<https://#{community.subdomain.name}/api/unsubscribe/#{subject}/#{token}>"
+  end
```
NeoVier, line 93: Maybe it would make more sense to just return the url (without the brackets), and include the brackets in the calling code, to be a bit more reusable

```suggestion
  def one_click_unsub(member, community, subject) do
    token = AuthToken.sign(member, "email")

    "https://#{community.subdomain.name}/api/unsubscribe/#{subject}/#{token}"
  end
```

NeoVier (approved):
Sorry for the delay 🏃‍♂️ 

I have a few comments, but it looks good overall!

NeoVier:
O botão de unsubscribe apareceu pra mim (enviando para um email do gmail, vendo pelo app Mail da Apple). Cliquei nele, apareceu uma caixa de confirmação, e quando eu aceitei não aconteceu nada (as opções de email continuam habilitadas). Tentei de novo, mas agora o botão não aparece mais. Talvez o botão não funcionou por não ser confiável ainda ou algo assim

lib/cambiatus_web/auth_token.ex
lucca65: If the page/token is only for changing email preferences, then I think its ok to have a big invalidation window. But if its for regular usage, it should be shorter for sure. Something around the week values. 

I'm not sure if Phoenix automatically updates the token validity based on usage (ie, if I keep using the app daily, it don't ever invalidate)

lib/cambiatus_web/email.ex
```diff
@@ -21,55 +25,93 @@ defmodule CambiatusWeb.Email do
...
-    |> from({"#{claim.action.objective.community.name} - Cambiatus", "no-reply@cambiatus.com"})
-    |> to(claim.claimer.email)
-    |> subject(gettext("Your claim was approved!"))
-    |> render_body("claim.html", %{claim: claim})
+    |> from({"#{community.name} - Cambiatus", Mailer.sender()})
+    |> to(claimer.email)
     |> set_language(claim)
+    |> subject(gettext("Your claim was approved!"))
+    |> header("List-Unsubscribe", one_click_unsub(claimer, community, "claim_notification"))
+    |> header("List-Unsubscribe-Post", "List-Unsubscribe=One-Click")
```
lucca65, line 52: maybe this and `set_language` could be a default header? Maybe with some function pattern matches to ensure every case (transfer, claim) is met

lib/cambiatus_web/email.ex
lucca65: +1, maybe we can actually create an issue to change this to a job queue, where we can configure sending limits. As we talked on Friday, I think this should be done separately, as this will be somewhat of a (very needed) refactor. 

This is a good sign. No serious production project lives on without some kind of background processing part!

lib/cambiatus_web/email.ex
lucca65: +1

lib/cambiatus_web/email.ex
```diff
@@ -21,55 +25,93 @@ defmodule CambiatusWeb.Email do
...
+    "https://#{community.subdomain.name}/unsubscribe?lang=#{language}&token=#{token}"
+  end
+
   def current_year, do: DateTime.utc_now() |> Date.year_of_era() |> Tuple.to_list() |> hd
 
   def format_date(date) do
     [date.day, date.month, date.year]
-    |> Enum.map(&to_string/1)
-    |> Enum.join("/")
+    |> Enum.map_join("/", &to_string/1)
```
lucca65, line 105: had no idea a `map_join` existed kkkk

lucca65 (changes requested):

lib/cambiatus_web/email.ex
MatheusBuss: Credo actually pointed it out to me, so I just followed :sweat_smile:

lib/cambiatus_web/email.ex
MatheusBuss: I agree with @NeoVier to create a function to compose emails, this way it's more standardized in code ans also easiser to create new emails. Maybe it could deal with everything that refers to community and recipient user.

lib/cambiatus_web/views/email_view.ex
MatheusBuss: Yeah, Phoenix gets really unhappy without this one.

MatheusBuss:
> If the page/token is only for changing email preferences, then I think its ok to have a big invalidation window. But if its for regular usage, it should be shorter for sure. Something around the week values.

This is something that could be done on #262. maybe we could have it be two weeks. Whats do you think @lucca65 and @NeoVier?

lib/cambiatus_web/auth_token.ex
MatheusBuss: This is something that could be done on https://github.com/cambiatus/backend/pull/262. maybe we could have it be two weeks. Whats do you think @lucca65 and @NeoVier?

lucca65:
thats ok for me!

lib/cambiatus_web/auth_token.ex
NeoVier: Sure, I think that's ok

lib/cambiatus_web/email.ex
MatheusBuss: Created #263

lib/cambiatus_web/email.ex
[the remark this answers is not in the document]
MatheusBuss: I'm looking into this. There is the [Cldr](https://hexdocs.pm/ex_cldr/1.6.4/readme.html) library that is meant to handle this type of localization. I think it's worth it to spend some time to get it working beacause in addition to formatting dates it can localize other data such as currencies and numbers. This then could be used by future issues, such as #198.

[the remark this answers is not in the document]
NeoVier: If you feel like this is going to take some bigger effort, it may be worth creating a new issue and dealing with it separately, since that's not really the focus of this PR, but that's your call 😄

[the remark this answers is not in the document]
MatheusBuss: Should be fine for this one. Already gonna test it on staging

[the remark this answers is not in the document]
MatheusBuss: image
Seems like it works

lucca65 (approved):
✨

Issue #148, opened 2021-04-21
Add subscriptions when something has to be written to the blockchain


zacck:
**Is your feature request related to a problem? Please describe.**
In a number of situtations our application has to write information the the blockchain then that has to be transferred to the backend and finally the front end can show it. 

Sometimes it's misleading to show a user stale information such as when an action is being created, redirecting back to the objective too soon would make i it look like the action creation failed

**Describe the solution you'd like**
We need to list the updates that go through the blockchain and make subscriptions for them so that we can have the frontend wait for their creation before  it can show a user some results 

**Describe alternatives you've considered**
None 

**Additional context**
Data needs to go to the blockchain then it can be broadcasted to the backend, this usually takes a few seconds and we should initiate waits for this 

@lucca65  I think a while ago you listed these scenarios in slack but I can't seem to find the message in question

PR #265, opened 2022-07-28
Fix/rich link empty bio

### LINKED ISSUE #264: Prevent a nil bio field to crash rich link rendering
Currently if a user's bio field is stored as `nil` it will be passed to the function `md_to_txt/1` in `lib/cambiatus_web/rich_link_view.exs` and crash it.

This behavior should be fixed.

MatheusBuss:
Closes #264 

Fill rich link description field when the user has no bio

### DIFF
--- a/lib/cambiatus_web/controllers/rich_link_controller.ex
+++ b/lib/cambiatus_web/controllers/rich_link_controller.ex
@@ -60,7 +60,7 @@ defmodule CambiatusWeb.RichLinkController do
            image: if(images != [], do: Map.get(List.first(images), :uri), else: @fallback_image),
            locale: get_language(language, creator),
            price: product.price,
-           currency: String.slice(community.symbol, 2, 7),
+           currency: String.split(community.symbol, ",") |> Enum.at(1),
            creator: Map.get(creator, :name) || creator.account
          }}
 
--- a/lib/cambiatus_web/controllers/rich_link_controller.ex
+++ b/lib/cambiatus_web/controllers/rich_link_controller.ex
@@ -88,10 +88,7 @@ defmodule CambiatusWeb.RichLinkController do
 
   def category_rich_link(category_id, community, language) do
     case Cambiatus.Shop.get_category(category_id) do
-      nil ->
-        {:error, "Category not found"}
-
-      category ->
+      {:ok, category} ->
         {:ok,
          %{
            type: :category,
@@ -101,6 +98,9 @@ defmodule CambiatusWeb.RichLinkController do
            image: category.icon_uri || category.image_uri || @fallback_image,
            locale: get_language(language, %{})
          }}
+
+      {:error, reason} ->
+        {:error, reason}
     end
   end
 
--- a/lib/cambiatus_web/controllers/rich_link_controller.ex
+++ b/lib/cambiatus_web/controllers/rich_link_controller.ex
@@ -53,6 +53,7 @@ defmodule CambiatusWeb.RichLinkController do
 
         {:ok,
          %{
+           type: :product,
            description: product.description,
            title: product.title,
            url: community.subdomain.name <> "/shop/#{product.id}",
@@ -73,7 +74,8 @@ defmodule CambiatusWeb.RichLinkController do
       {:ok, user} ->
         {:ok,
          %{
-           description: user.bio || (user.name || user.account) <> " is on " <> community.name,
+           type: :user,
+           description: user.bio,
            title: user.name || user.account,
            url: community.subdomain.name <> "/profile/#{user.account}",
            image: user.avatar || @fallback_image,
@@ -93,6 +95,7 @@ defmodule CambiatusWeb.RichLinkController do
       category ->
         {:ok,
          %{
+           type: :category,
            description: category.meta_description || category.description,
            title: category.meta_title || category.name,
            url: community.subdomain.name <> "/shop/categories/#{category.slug}-#{category.id}",
@@ -105,6 +108,7 @@ defmodule CambiatusWeb.RichLinkController do
   def community_rich_link(community, language) do
     {:ok,
      %{
+       type: :community,
        description: community.description,
        title: community.name,
        url: community.subdomain.name,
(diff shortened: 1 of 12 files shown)


lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -83,40 +87,36 @@ defmodule CambiatusWeb.RichLinkController do
...
-       }}
-    else
-      {:error, reason} ->
-        {:error, reason}
-    end
-  end
+  def category_rich_link(category_id, community, language) do
+    case Cambiatus.Shop.get_category(category_id) do
+      nil ->
+        {:error, "Category not found"}
```
MatheusBuss, line 93: @lucca65 maybe we should make `Cambiatus.Shop.get_category` consistent with other get functions, as in make it return either `{:ok, category}` or `{:error, reason}`

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -83,40 +87,36 @@ defmodule CambiatusWeb.RichLinkController do
...
+    {:ok,
+     %{
+       type: :community,
+       description: community.description,
+       title: community.name,
+       url: community.subdomain.name,
+       image: community.logo,
+       locale: get_language(language, %{})
+     }}
+  end
```
MatheusBuss, line 118: Since I'm revisiting this, it could be a good idea to set the community language properly. We could make it get the community creator language or have some way to set a default language for every community.

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -44,37 +46,39 @@ defmodule CambiatusWeb.RichLinkController do
...
+        {:ok,
+         %{
+           type: :product,
+           description: product.description,
+           title: product.title,
+           url: community.subdomain.name <> "/shop/#{product.id}",
+           image: if(images != [], do: Map.get(List.first(images), :uri), else: @fallback_image),
+           locale: get_language(language, creator),
+           price: product.price,
+           currency: String.slice(community.symbol, 2, 7),
```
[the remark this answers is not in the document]
NeoVier: The current implementation actually wouldn't work for this community I have: https://precisionthree.staging.cambiatus.io/dashboard. The symbol is 7 characters long (`3,THREEPR`)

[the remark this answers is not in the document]
MatheusBuss: This makes a lot of sense actually. I thought we had some limit to the symbol string. Good catch!

NeoVier (approved):
✨

lib/cambiatus_web/controllers/rich_link_controller.ex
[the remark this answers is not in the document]
NeoVier: We do have some limits, I think it must be 3 <= length <= 7. But we don't need to worry about that with the `String.split` method

lib/cambiatus_web/controllers/rich_link_controller.ex
lucca65: +1, this is not consistent at all in our contexts. I do think we need to aim for consistency

lib/cambiatus_web/controllers/rich_link_controller.ex
lucca65: One thing to keep in mind is that community creation happens on the blockchain, and its synced up to our database. This means that we don't have the creators language at our context, unless we modify the contract to add this extra information. IMO, its not worth the trouble, at least for now. 

I mean, at some point we will need to walk you through contract modifications and how to respond to that in our backend. We have #148 that seems like a nice spot for this exercise

lib/cambiatus_web/controllers/rich_link_controller.ex
```diff
@@ -7,31 +7,33 @@ defmodule CambiatusWeb.RichLinkController do
...
 
-  alias CambiatusWeb.Resolvers.{Accounts, Commune, Shop}
+  alias CambiatusWeb.Resolvers.{Accounts, Shop}
   alias Cambiatus.Repo
 
   @fallback_image "https://cambiatus-uploads.s3.amazonaws.com/cambiatus-uploads/b214c106482a46ad89f3272761d3f5b5"
 
   def rich_link(conn, params) do
     language = get_req_header(conn, "accept-language")
+    context = conn.private[:absinthe][:context]
```
lucca65, line 17: This don't make sense, absinthe context is relevant when we receive GraphQL over HTTP requests. Controllers don't handle that. 

Remember that we have two APIs on our backend today: GraphQL and Rest. We should consider them completely separate.

lucca65 (changes requested):
I think we got the wrong concept here on this issue bro. We don't have community information over HTTP Rest, only given our graphql Stuff. 

Think about it. The GraphQL client is the frontend, but the user of our Rich Link is the scrapping bots and Apps previewers. They don't do that using GraphQL, but rather plain json/http. 

Unless I'm missing something this is not going to fix it for our rich link use cases

lib/cambiatus_web/controllers/rich_link_controller.ex
MatheusBuss: Oops, forgot about that, sorry. It was just bothering me that the community is already stored in memory and the controller had to fetch it again.

I did some refactoring to place the current community also into conn itself, this way it's not reliant on GraphQL. But this does imply in having the same information passing around the app twice on every call.

lib/cambiatus_web/controllers/rich_link_controller.ex
MatheusBuss: My idea for this case was to get the user from the creator field in the Community struct and use this language. It's not perfect, but it's something

lib/cambiatus_web/controllers/rich_link_controller.ex
MatheusBuss: :heavy_check_mark: Done

lucca65 (approved):
✅

PR #272, opened 2022-08-04
Create shop orders

### LINKED ISSUE #198: Create Shop orders

MatheusBuss:
Closes #198 

Create shop orders abstraction

### DIFF
--- a/lib/cambiatus/orders.ex
+++ b/lib/cambiatus/orders.ex
@@ -1,0 +1,321 @@
+defmodule Cambiatus.Orders do
+  @moduledoc """
+  The Orders context.
+  """
+
+  import Ecto.Query, warn: false
+  alias Cambiatus.Repo
+
+  alias Cambiatus.Orders.Order
+
+  @doc """
+  Returns the list of orders.
+
+  ## Examples
+
+      iex> list_orders()
+      [%Order{}, ...]
+
+  """
+  def list_orders do
+    Repo.all(Order)
+  end
+
+  @doc """
+  Gets a single order.
+
+  Returns {:error, "No order exists with the id: {id}} if the Order does not exist.
+
+  ## Examples
+
+      iex> get_order(123)
+      {:ok, %Order{}}
+
+      iex> get_order(456)
+      {:error, "Could not find order with id: {id}}
+
+  """
+  def get_order(id) do
+    case Repo.get(Order, id) do
+      nil ->
+        {:error, "No order exists with the id: #{id}"}
+
+      val ->
+        {:ok, val}
+    end
+  end
+
+  @doc """
+  Gets a single order.
+
+  Raises `Ecto.NoResultsError` if the Order does not exist.
+
+  ## Examples
+
+  iex> get_order!(123)
+  %Order{}
+
+  iex> get_order!(456)
+  ** (Ecto.NoResultsError)
+
+  """
+  def get_order!(id), do: Repo.get!(Order, id)
+
+  @doc """
+  Gets a single order.
+
+  Returns {:error, "No order exists with the id: {id}} if the Order does not exist.
+
+  ## Examples
+
+      iex> get_order(123)
+      {:ok, %Order{}}
+
+      iex> get_order(456)
+      {:error, "Could not find order with id: {id}}
+
+  """
+
+  # TODO: Add current_user to query
+  def get_shopping_cart() do
+    Order
+    |> where([o], o.status == "cart")
+    |> Repo.all()
+    |> case do
+      [] ->
+        {:error, "User has no shopping cart"}
+
+      [value] ->
+        {:ok, value}
+
+      _ ->
+        {:error, "User has more than one shopping cart"}
+    end
+  end
+
+  @doc """
+  Creates an order.
+
+  ## Examples
+
+      iex> create_order(%{field: value})
+      {:ok, %Order{}}
+
+      iex> create_order(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def create_order(attrs \\ %{}) do
+    %Order{}
+    |> Order.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  @doc """
+  Creates an order.
+
+  ## Examples
+
+      iex> create_order!(%{field: value})
+      %Order{}
+
+      iex> create_order!(%{field: bad_value})
+      Error
+
+  """
+  def create_order!(attrs \\ %{status: "cart", payment_method: "eos", total: 0}) do
+    %Order{}
+    |> Order.changeset(attrs)
+    |> Repo.insert!()
+  end
+
+  @doc """
+  Updates a order.
+
+  ## Examples
+
+      iex> update_order(order, %{field: new_value})
+      {:ok, %Order{}}
+
+      iex> update_order(order, %{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def update_order(%Order{} = order, attrs) do
+    order
+    |> Order.changeset(attrs)
+    |> Repo.update()
+  end
+
+  @doc """
+  Deletes a order.
+
+  ## Examples
+
+      iex> delete_order(order)
+      {:ok, %Order{}}
+
+      iex> delete_order(order)
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def delete_order(%Order{} = order) do
+    Repo.delete(order)
+  end
+
+  @doc """
+  Returns an `%Ecto.Changeset{}` for tracking order changes.
+
+  ## Examples
+
+      iex> change_order(order)
+      %Ecto.Changeset{data: %Order{}}
+
+  """
+  def change_order(%Order{} = order, attrs \\ %{}) do
+    Order.changeset(order, attrs)
+  end
+
+  alias Cambiatus.Orders.Item
+
+  @doc """
+  Returns the list of items.
+
+  ## Examples
+
+      iex> list_items()
+      [%Item{}, ...]
+
+  """
+  def list_items do
+    Repo.all(Item)
+  end
+
+  @doc """
+  Gets a single item.
+
+  Returns {:error, "No item exists with the id: {id}} if the Item does not exist.
+
+  ## Examples
+
+      iex> get_item(123)
+      {:ok, %Item{}}
+
+      iex> get_item(456)
+      {:error, "Could not find item with id: {id}}
+
+  """
+  def get_item(id) do
+    case Repo.get(Item, id) do
+      nil ->
+        {:error, "No item exists with the id: #{id}"}
+
+      val ->
+        {:ok, val}
+    end
+  end
+
+  @doc """
+  Gets a single item.
+
+  Raises `Ecto.NoResultsError` if the Item does not exist.
+
+  ## Examples
+
+      iex> get_item!(123)
+      %Item{}
+
+      iex> get_item!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_item!(id), do: Repo.get!(Item, id)
+
+  @doc """
+  Creates a item.
+
+  ## Examples
+
+      iex> create_item(%{field: value})
+      {:ok, %Item{}}
+
+      iex> create_item(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def create_item(attrs \\ %{})
+
+  def create_item(%{order_id: id} = attrs) do
+    %Item{}
+    |> Item.changeset(attrs)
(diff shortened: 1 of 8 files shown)


MatheusBuss:
@lucca65 criei um contexto novo pra orders. Mas acho que deveria ter jogado ele pra dentro do shop mesmo. Acho que fica mais organizado. Daí teria que reconstruir o arquivo `lib/cambiatus/shop/order.ex`.

lucca65:
+1 bro, vamos manter td no shop!

lib/cambiatus/orders.ex
```diff
@@ -77,9 +78,11 @@ defmodule Cambiatus.Orders do
   """
 
   # TODO: Add current_user to query
-  def get_shopping_cart() do
+  def get_shopping_cart(buyer) do
     Order
     |> where([o], o.status == "cart")
+    |> join(:left, [o], b in assoc(o, :buyer))
+    |> where([o, b], b.account == ^buyer.account)
```
lucca65, line 85: can't we simply preload the buyer? Something like this:

```suggestion
    |> Repo.preload(:buyer)
```

lib/cambiatus/orders.ex
```diff
@@ -130,8 +133,8 @@ defmodule Cambiatus.Orders do
   end
 
   @doc """
-  Updates a order.
-
+  Updates an order.
```
lucca65, line 136: 🇬🇧

lib/cambiatus/orders/item.ex
```diff
@@ -23,12 +25,28 @@ defmodule Cambiatus.Orders.Item do
...
     item
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
   end
+
+  def validate_item_stock(%{id: id} = changeset) do
+    with {:ok, item} <- Orders.get_item(id),
+         item <- Repo.preload(item, :product),
+         product <- Map.get(item, :product) do
+      case product.track_stock do
```
lucca65, line 41: maybe an `if`? case for Boolean values don't make a lot of sense

lib/cambiatus/orders/order.ex
```diff
@@ -18,12 +23,21 @@ defmodule Cambiatus.Orders.Order do
...
     order
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
+    |> validate_checkout()
+  end
+
+  def validate_checkout(changeset) do
+    with order_id <- get_field(changeset, :id),
+         {:ok, order} <- Orders.get_order(order_id) do
+      if Map.get(order, :status) == "cart" and get_field(changeset, :status) == "checkout" do
```
lucca65, line 39: This could be piped

```suggestion
order_status = 
    changeset 
    |> get_field(:id)
    |> Orders.get() 
    |> Map.get(:status)
    
    if Enum.member(["cart", "checkout"], order_status) do
      . . . 
```

lucca65:
I know its a draft, some general directions 😄

lib/cambiatus/orders.ex
MatheusBuss: :heavy_check_mark:

lib/cambiatus/orders.ex
MatheusBuss: I don't think so. Because this is getting an order where the status is `cart` and belongs to the current user. Since we're getting the order from these SQL statements I don't believe it's possible to preload the buyer.

lib/cambiatus/orders/item.ex
MatheusBuss: I've changed this one a bit. Not sure which would be better: using two nested `if`'s or a case that returns a tuple with two booleans. Something like this:
```elixir
      if product.track_stock do
        if product.units >= item.units do
          changeset
        else
          add_error(changeset, :units, "Not enough product in stock")

      else
        changeset
      end
```
OR
```elixir
      case {product.track_stock, product.units >= item.units} do
        {true, true} ->
          changeset

        {true, false} ->
          add_error(changeset, :units, "Not enough product in stock")

        {false, _} ->
          changeset
      end
```

MatheusBuss: I've opted for the second one, since I don't think the nested `if`'s are an example of great readability

lib/cambiatus/orders/order.ex
MatheusBuss: The intent of this one is to check if the order stored in the db has the status `cart` and the incoming changeset is changing this status to `checkout`. Then we should make some more validations to ensure that everything is valid before the user is has the chance to make a payment.

This has also changed a bit. If you want to make another review it'll be very welcome.

lib/cambiatus/orders/item.ex
NeoVier: How about

```elixir
if product.track_stock && product.units < item.units do # product.units < item.units == not(product.units >= item.units)
  add_error(...)
else
  changeset
```

MatheusBuss: That would be a way better solution in every regard, yes hahaha

lucca65: lol! guess that is something that happens between programming styles, imperative will usually go to operators like `&&` `||` and functional will look for tuples values to match on. 

In this case the `if` makes more sense, but to be honest both solutions (nested if and pattern match tuple'd values) works well too.

The only one I think its weird is using case to match boolean values 😀

lib/cambiatus/shop/order.ex
```diff
@@ -1,23 +1,80 @@
...
   schema "orders" do
-    field(:amount, :float)
-    field(:units, :integer)
+    field(:payment_method, Ecto.Enum,
+      values: [:paypal, :bitcoin, :ethereum, :eos],
+      default: :paypal
+    )
+
+    field(:total, :float)
+    field(:status, :string)
```
lucca65, line 22: Maybe we need to review possible states bro. Can you implement this as an `Ecto.Enum`? This way we can talk more about control flow and state changes. I recommend you take a look at state machines, as this will be a nice way to think about this abstractions. A gist of it could be something like

**states**: `created`, `processed`, `approved`, `completed`
**actions**: *new_draft* (makes a new one), approve (`created` → `approved` or `processed` → `approved`, pay (no state change), archive (any → `completed`

lucca65: this is just an example on how to approach an order state as its heavily implied that it will be treated different depending on it. This means we will have specific actions that will change state. Makes easier to understand and admin

priv/repo/migrations/20220803194923_create_orders.exs
```diff
@@ -0,0 +1,26 @@
...
+
+      add(:payment_method, :payment_method,
+        default: "eos",
+        null: false,
+        comment: "Payment method used, typed with the integrations we got"
+      )
+
+      add(:total, :float, comment: "Order total")
+
+      # TODO: Elaborate shipping and status
```
lucca65, line 16: Shipping will be an entity that we will attach on later, we can do a preview of states but don't worry about tables/logic for shipping

priv/repo/migrations/20220803140956_rename_orders_to_items.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Cambiatus.Repo.Migrations.RenameOrdersToItems do
+  use Ecto.Migration
+
+  def change do
+    rename(table(:orders), to: table(:items))
```
lucca65, line 5: why items?

lib/cambiatus/shop/order.ex
MatheusBuss: I was thinking about elaborating states. If you want we can have a more thorough discussion about states on a meeting of some kind, maybe even with other teams to get the right flow.

But I was thinking something like this:
1 - **Cart**: There's no more than one order with this status for each user
2 - **Checkout**:Validate everything that was put on the cart, since availability may have changed since adding to cart
3 - **Awaiting payment confirmation**: This should be a brief state, just to ensure that the items on checkout have been payed for but we're making sure that everything went through
4 - **Delivering**: This could change name or encompass other steps. Some orders may need preparation after payment but before beign archive/completed
5 - **Completed**: End step for every order, maybe there should be another end step if something goes wrong along the way

priv/repo/migrations/20220803140956_rename_orders_to_items.exs
MatheusBuss: Since every order until now has been a single product purchase it is probably easier to treat them as items, since most of the information is being contained within the items table. Then we can create the related orders with some info on the buyer.

We can always do it the other way as well, shouldn't be too different.

lucca65: hmm, Idk. We will still need to major abstraction of orders, which is the main one, that can start with a product_id as a reference to the product and later on can be migrate into an intermediary table `order_product_id`. 

About the majority of the information, it should be on orders, not on items.

The order table could be scoped like this:

- id
- product_id
- from
- to
- amount
- units
- community
- blockchainstuff

We could refactor it like this:

- id
- product_id (from can be inferred from product, the same with community)
- buyer
- amount (can be different from product due to discounts cupons and other stuff so we should keep it)
- units
- timestamps

Later we could trim it even further by adding a middle table

**Order**
- id
- product_id
- buyer
- timestamps

**Order_items**
- order_id (Primary key)
- product_id (Primary key)

**Order_payments**
- id
- type (enum with something like btc/paypal/venmo/eth/etc)
- status (created/confirmed/rejected/etc)
- amount 
- price_modifier_id (maybe a certain payment method is cheaper or more expensive than others. Crypto has no fees, as PayPal do)

Then we could go to other attached tables **order_shipping**, **order_refunds**, etc

We should try to go for the higher level abstraction first, then add the auxiliary items. It will make our migration harder, but we can go in small steps as described in the beginning of the comment

lib/cambiatus/shop/order.ex
lucca65: yeah, you are totally right, nailing the states will go a long way for us, migrating types is a pain in the ass so the better we design our states, the best it is.

I haven't thought about adding cart as an state, this is a good idea. I would add other auxiliary states as well. Cancelled by vendor, cancelled by buyer (maybe by not paying during the accepted window?) payment denied and finally payment incomplete (this is a must as the main motivation to building orders is to allow multiple payments on the same order)


+1 on scheduling a meeting. We should do it between ourselves first to build a first draft and then present it to the rest of the team to gather feedback

lib/cambiatus/orders/order.ex
lucca65: This cart I've added a few comments about state machines, but we should leverage its structure to make sure its valid, instead of relying on manual verification. 

If an order is on a certain state, we should trust our backend ingested it properly so its correct to have this state. This is an old article and probably outdated, but it contain precious concepts about this type of abstraction: https://medium.com/@joaomdmoura/state-machine-in-elixir-with-machinery-8ee6f9def2da

PR #273, opened 2022-08-06
Fix two wrong instances of product_category mutations

lucca65:
Closes N/A

- Adjust the way we relate product and product categories
- Add extra info (community_id and creator_id) to product when creating
- Make sure we can create new products with categories related
- Make sure we can change categories of an existing product that already had categories 

- `mix test`

### DIFF
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -1,5 +1,4 @@
-# 5 minutes
-ExUnit.start(timeout: 5 * 60_000, exclude: [:skip])
+ExUnit.start(timeout: 5 * 60_000, exclude: [:skip], capture_log: true)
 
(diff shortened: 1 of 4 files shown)


NeoVier (approved):
Works like a charm ✨

test/test_helper.exs
```diff
@@ -1,5 +1,4 @@
-# 5 minutes
-ExUnit.start(timeout: 5 * 60_000, exclude: [:skip])
+ExUnit.start(timeout: 5 * 60_000, exclude: [:skip], capture_log: true)
```
MatheusBuss, line 1: Nice!

MatheusBuss (approved):
That was fast :zap:

PR #295, opened 2022-09-02
Fix/fix monthly digest workers

### LINKED ISSUE #294: Refactor monthly digest email worker
When we implemented generalized workers for email on #276 we left behind a bug.

We spawn a worker for an entire community. Which means that, if a email for a single user in that community fails, the entire worker fails. This also can cause some users to receive the same message multiple times, since the worker will retry sending emails to the entire community up to 3 times.

We should change this behavior to spawn a worker for every member, instead of a worker for every community.

MatheusBuss:
Closes #294 

Refactor Monthly Digest Worker to make it so that an individual worker is responsible for an individual user, instead of a single worker being responsible for the entire community.

Run `mix test`

### DIFF
--- a/lib/cambiatus/workers/monthly_digest_worker.ex
+++ b/lib/cambiatus/workers/monthly_digest_worker.ex
@@ -12,8 +12,6 @@ defmodule Cambiatus.Workers.MonthlyDigestWorker do
   alias Ecto.Multi
 
   def perform(_) do
-    multi = Multi.new()
-
     transaction =
       Community
       |> Community.with_news_enabled()
@@ -24,7 +22,7 @@ defmodule Cambiatus.Workers.MonthlyDigestWorker do
         :subdomain
       ])
       |> Enum.filter(&Enum.any?(&1.news))
-      |> Enum.reduce(multi, fn community, multi ->
+      |> Enum.reduce(Multi.new(), fn community, multi ->
         Enum.reduce(community.members, multi, fn member, multi ->
           Oban.insert(
             multi,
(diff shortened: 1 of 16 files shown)


lib/cambiatus/workers/monthly_digest_worker.ex
```diff
@@ -1,23 +1,39 @@
...
+      Community
+      |> Community.with_news_enabled()
+      |> Repo.all()
+      |> Repo.preload([
+        [news: News.last_thirty_days()],
+        [members: User.accept_digest()],
+        :subdomain
+      ])
+      |> Enum.filter(&Enum.any?(&1.news))
+      |> Enum.reduce(multi, fn community, multi ->
```
NeoVier, line 27: I think this would be a bit clearer - we wouldn't have so many `multi` around:

```suggestion
      |> Enum.reduce(Multi.new(), fn community, multi ->
```

lib/cambiatus_web/email.ex
```diff
@@ -41,14 +42,15 @@ defmodule CambiatusWeb.Email do
...
-  # input is a community with preloaded news with less than 30 days and members with active digest
-  def monthly_digest(community) do
-    Enum.each(community.members, fn member ->
-      compose_email_headers(member, community)
-      |> subject("#{community.name} - " <> gettext("Community News"))
-      |> render_body("monthly_digest.html", render_params(member, community))
-      |> Mailer.deliver()
-    end)
+  # input is a community and a user
+  # community and memebr filtering is done by the MonthlyDigestWorker
```
NeoVier, line 46:
```suggestion
  # community and member filtering is done by the MonthlyDigestWorker
```

NeoVier (approved):

lib/cambiatus_web/email.ex
MatheusBuss: Oops! Nice catch

lib/cambiatus/workers/monthly_digest_worker.ex
MatheusBuss: More concise indeed

lucca65 (approved):
✅

PR #303, opened 2022-12-08
Add community roles

lucca65:
Closes N/A

Simple PR that adds community roles. Getting up to speed after a break from coding

- `mix check`
- `mix test`


.github/workflows/test.yml
```diff
@@ -25,27 +28,16 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: '24.2'
+          elixir-version: '1.14.2'
+
+      - run: sudo apt -y install exiftool
+      - run: mix deps.get
+      - run: mix format --check-formatted    
+      # - run: mix do compile --warnings-as-errors --all warnings, format --check-formatted, credo --strict --only warning
```
MatheusBuss, line 40: Shouldn't this be *not* commented?

Maybe better yet would be to remove `format --check-formatted` (since it's on the line above) an run credo before compile.

```suggestion
      - run: mix credo --strict --only warning
      - run: mix compile --warnings-as-errors --all warnings
```

MatheusBuss (commented):
Nice idea to categorize info on communte types!

MatheusBuss (approved):

